PR #220, opened 2015-03-16
Delete unnecessary checks before calls of the function "free".

elfring:
[The function "free"](http://stackoverflow.com/questions/18775608/free-a-null-pointer-anyway-or-check-first "Free a null pointer anyway or check first?") is documented in the way that no action shall occur for a passed null pointer.
Redundant safety checks can be avoided.


allinurl:
Thanks! This has been merged.

PR #251, opened 2015-05-21
Installation instructions for OpenBSD

fcambus:


allinurl:
Merged. Thanks for the update!

PR #254, opened 2015-05-27
add sortable.js and styling

macbookandrew:
Something I thought might be helpful: a JS sorting plugin to sort the HTML output tables.

I’m not at all familiar with programming in C and couldn’t get the JS working when embedded in the file, so it’s loading from a CDN. The styles, however, are embedded in the HTML output.

One other note: columns are sorted by the data in them, but that’s obviously a problem for the bandwidth columns; adding a `data-value=""` attribute with the number of bytes to each affected `<td>` should handle sorting correctly (reference: “Sorting options” on this page: http://github.hubspot.com/sortable/api/options/). Again, I’m not familiar with C programming and don’t have the faintest idea where or how to add that attribute.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -422,6 +422,8 @@ print_html_header (FILE * fp, char *now)
 
   fprintf (fp, "</script>\n");
 
+  fprintf (fp, "<script type=\"text/javascript\" src=\"https://cdnjs.cloudflare.com/ajax/libs/sortable/0.6.0/js/sortable.min.js \"></script>\n");
+
   fprintf (fp, "<style type=\"text/css\">");
   fprintf (fp,
   "html {"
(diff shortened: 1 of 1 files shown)


allinurl:
Thanks. I was playing with this, however, are you able to sort all tables? it's only sorting some tables for me though. I'm also curious if there is an option for colspans in the case of the hosts panel. i.e., using `-a`.

macbookandrew:
No…I hadn’t noticed that and can’t seem to get it working for all tables…not sure what the problem is. Oh well…

macbookandrew:
So it’s working all the way through a different test file…there must be something in certain reports that is breaking it.

allinurl:
I'm curious, did you run it with `-a` to enable a list of user agents per host?

macbookandrew:
No, I didn’t use the -a option. 


Andrew Minion / Web & Print Designer
andrew@andrewrminion.com / 704-769-0625

macbookandrew:
I just tried with the `-a` option and sortable really goofs up that table :smile: Probably best to disable the `data-sortable` attribute from being added to that table when `-a` is used, but that’s beyond my C programming capabilities.

screen shot 2015-06-11 at 8 30 24 am

allinurl:
I'll take a look and see what can be done on this table, as well as those that have aggregated rows, e.g., browsers, os. Thanks anyway.

PR #278, opened 2015-08-06
Update util.c

mynameiscfed:
Updated error codes to support CloudFlare custom error codes.


allinurl:
This has been merged. Thanks!

PR #287, opened 2015-08-28
Don't override CFLAGS

cicku:


allinurl:
Merged. Thanks for the contribution!

PR #288, opened 2015-08-28
Drop math library linking flag

cicku:
I haven't found math.h #include in sources, probably a cleanup mess from the past?

Also if you want to use libm, please use configure.ac:

```
AC_CHECK_LIB(m, THE_MATH_FUNCTION_GOACCESS_USES_BUT_NOT_SHOWN_IN_GIT_SO_I_DONT_KNOW_YET)
```


allinurl:
A while back it was using natural logarithm. Merged, thanks!

PR #289, opened 2015-08-28
Refine configure output

cicku:
This patch outputs ```$LDFLAGS``` with ```$LIBS``` to detect if goaccess is properly linked.


allinurl:
Merged, thanks!

PR #337, opened 2015-11-24
Virtualmin Support Update goaccess.conf

CanadianJeff:
Virtualmin uses a format that is not NCSA Standard.


allinurl:
Merged. Thanks!

PR #349, opened 2016-01-01
Correctly link against libtinfo

SpiderX:
When ncurses is built with --with-termlib=tinfo option then there are two libraries - libtinfo (which contains terminal related functions) and libncurses (rest). We should correctly link against libtinfo in such case.


allinurl:
Merged. Thank you!

Issue #192, opened 2014-12-05
Please add a javascript table sort


pjsg:
Using something like jquery tablesorter would allow the various tables to be sorted by the user clicking in the browser.

Thanks

2vek:
:+1: Would be nice to have.
Since, we already have expanding tables, first thing I tried to do was sorting table by clicking table header.

allinurl:
@2vek I'll look into it now that I'm working on the Bootstrapify output #223. Thanks for the reminder!

allinurl:
This has been implemented and deployed in [version 1.0](https://goaccess.io/). Thanks.

PR #375, opened 2016-02-21
HTML - sorting tables by clicking on table header

osterik:
Hello!

Thist is fast & dirty way for sorting tables by clicking on table header (issue https://github.com/allinurl/goaccess/issues/192)

todo:
1) integrate jquery & tablesorter into html (?).
I doubt that it is a good idea, because of increasing report file size, but to use external resources (CDN) for internal purporses also does not seem reasonable.

2) add sorting markers (up-down triangles).

3) resolve issue with disabling bandwith for correct sorting.

I'm very begiiner in modern front-end development, hope community can improve my solution.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -465,8 +465,12 @@ print_html_header (FILE * fp, char *now)
   fprintf (fp, "<meta name='robots' content='noindex, nofollow' />");
   print_html_title(fp, now);
 
+  fprintf (fp, "<!-- load jQuery and tablesorter scripts for sorting tables -->\n");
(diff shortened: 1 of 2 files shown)


allinurl:
@osterik I'll have to hold back on this since the plan is to have this implemented natively in v1.0 (which I'm currently working on). Regarding your points:

1. The plan is to restructure a bit the tables, so a basic table sorter should do the job without needing to add an extra dependency. Using a CDN becomes a privacy issue for some users.

2. We'll make sure to have these markers.

3. Will be addressed as well in v1.0

Thanks very much for the contribution though.

PR #376, opened 2016-02-22
Fix: bug if panel does not exist

d3f3kt:
If a panel does not exist in AppData but is set in AppUIData, JavaScript will throw a fatal error.
For example this happens if no virtual hosts are parsed.


allinurl:
Merged. Thank you!

PR #380, opened 2016-02-22
Update pagination

d3f3kt:
* Don't show pagination if it isn't necessary
* Diasable pagination buttons if first or last page is reached
* Fix a small bug


allinurl:
Merged. Thank you!

PR #381, opened 2016-02-22
Update Graphs

d3f3kt:
1) Add tooltip value formater
2) Use d3.format("d") to format y axis

====== 
The main change was to implement the tooltip value formater. 
But I also changed the y axis formater. @allinurl You have to decide if you want to use the d3.format("s") to format the y axis data. In my opinion it's much nicer to display 550M instead of 55000000.


allinurl:
Merged. Thank you!

I agree with you, the human readable format is the way to go. The update looks good :+1:

PR #455, opened 2016-06-30
Change grid row number start with 1

charpty:
Hi, I think number start with the zero is the programmer's habit not the customer's.

Signed-off-by: Bo Cai <charpty@gmail.com>


allinurl:
Merged. Thanks!

PR #463, opened 2016-07-06
Add note about OpenIndiana to README

xen0l:
I pushed a package into OpenIndiana repositories and it should appear in repos in few hours. This PR adds a note how to install goaccess on OpenIndiana.


allinurl:
Merged. Thanks!

PR #521, opened 2016-09-18
browser: classify google cloud clients

ishitatsuyuki:
For example, gsutil is based on Python but not a crawler.

### DIFF
--- a/src/browsers.c
+++ b/src/browsers.c
@@ -86,6 +86,7 @@ static const char *browsers[][2] = {
   {"Epiphany", "Others"},
   {"Firebird", "Others"},
   {"Galeon", "Others"},
+  {"google-cloud-sdk", "Others"},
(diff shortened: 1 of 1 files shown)


allinurl:
Thank you for this PR. Would you be able to post the log line containing this user agent?

ishitatsuyuki:
Since you're supporting Google Cloud Storage file, lines like this is not common since every admin operation is also logged...
`"time","ip","1","","GET","/storage/v1/b/path","200","0","0","167000","www.googleapis.com","","apitools gsutil/4.21 Python/2.7.12 (linux2) google-cloud-sdk/126.0.0 analytics/disabled","request_id","storage.objects.get","bucket","file"`

allinurl:
Thanks for posting this! It's been merged!

PR #579, opened 2016-11-29
Use bucket name as virtualhost for Amazon S3

rib3:
Use bucket name as virtualhost for Amazon S3.

This isn't as valuable as the request's virtualhost, but is still useful when analyzing aggregated logs.

[S3 logs](http://docs.aws.amazon.com/AmazonS3/latest/dev/LogFormat.html) unfortunately do not include virtualhost.

I also removed the whitespace padding before the comment to follow the pattern set by the `Cloud Storage` line.


allinurl:
Merged. Thank you!

PR #633, opened 2017-01-23
Added sanity check for db-path config variable

coderonline:
Just a small change, but it took me quite a while to figure out, why my real time html wasn't updated any more. The reason turned out to be a wrong setting for db-path: A file instead of a directory. Goaccess started without complaining, could even read the database, but not update it. So I wrote this little 'never trust in users input'-patch.


allinurl:
Awesome, thanks for catching that and sending the PR. I'll merge this and add the relevant header.

PR #639, opened 2017-01-25
Add more feed readers

da2x:
These are the feed reader services witt the most hits on my site not recongized by GoAccess.


allinurl:
Merged. Thank you!

PR #651, opened 2017-02-06
Docker support

### LINKED ISSUE #520: GoAccess Dockerfile
Just found this tool, looks brilliant!

I was hoping to use this along with Caddy to watch the access logs in my docker-compose stack. I found there was a couple very outdated third party implementations.

https://github.com/diyan/goaccess-docker
https://github.com/joelchen/goaccess

Both of these are pre-1.0.0. I was hoping for an official Docker image with support for running and watching the Caddy logs in realtime.

Thanks

julianxhokaxhiu:
I like this project and I was surprised that it doesn't come with Docker support.

Therefore I made a Dockerfile for you, which can be used on Docker Hub to Automatically Build it every time you commit or you release a new version.

At the moment it depends on Alpine package manager to fetch the latest version available, but would be nice to enhance the Docker by fetching the binary from the Releases tab here in Github.

Anyway, feel free to edit parts where you feel comfortable. Build is already tested ( as you can see here https://hub.docker.com/r/julianxhokaxhiu/goaccess/builds/ ).

Fixes #520

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,64 @@
+FROM alpine:3.4
+LABEL maintainer "Julian Xhokaxhiu <info@julianxhokaxhiu.com>"
+
+# Environment variables
+#######################
+
+ENV DATA_DIR /srv/data
+
+# Configurable environment variables
+####################################
+
+# Copy required files and fix permissions
+#########################################
+
+COPY Dockerfile_src/* /opt/
+
+# Create missing directories
+############################
+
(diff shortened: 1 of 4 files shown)


allinurl:
Thanks a lot for this PR. It will be pushed out in upcoming release!

allinurl:
Just for the record, #660 added the ability to daemonize the process using `--daemonize` along with `--real-time-html`.

julianxhokaxhiu:
Nice :) well done!

PR #653, opened 2017-02-07
Use the latest Alpine image

julianxhokaxhiu:
No reasons to hold us back.

We win also 1MB of free space :)


allinurl:
Merged. Thank you!

PR #694, opened 2017-03-21
Parse user configuration first when available (fixes #693)

### LINKED ISSUE #693: Should probe ~/.goaccessrc first
By default, goaccess tries to read the global configuration file, and only load the user configuration file if that's missing. This is not how normal *nix utility behave. On main distributions, the global configuration file is generally provided as a guideline, meaning that ~/.goaccessrc will be never scanned.

It would be more user-friendly to switch the probing order to ~/.goaccessrc -> /syspath/goaccess.conf

wavexx:
Probe for user's goaccessrc first, then fallback to system default.

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -129,17 +129,16 @@ get_config_file_path (void)
     return rpath;
   }
 
-  /* otherwise, fallback to global config file, or
-   * attempt to use the user's config file */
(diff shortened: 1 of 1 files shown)


allinurl:
Merged. Thank you!

Issue #701, opened 2017-03-24
Add support for multiple languages


NengWan:
the web interface is so cool !   can I change the display language?

allinurl:
Currently there are no other languages available. However, this is coming! It will probably need to be implemented using [`gettext`](https://www.gnu.org/software/gettext/)

There are a couple of pull requests ([**French**](https://github.com/allinurl/goaccess/pull/453) and [**Chinese**](https://github.com/allinurl/goaccess/pull/457)) that I need to merge and add gettext support to them.

allinurl:
This has been implemented. As of this post, French and Spanish have been added. If you would like to collaborate in adding a different language, please see take a look at [**this post**](https://github.com/allinurl/goaccess/issues/858#issuecomment-320257517) and I'll be happy to merge it for the upcoming version.

tellustheguru:
I have translates GoAccess to Swedish, how do I send it to you?

allinurl:
@tellustheguru Thanks for offering to translate it to Swedish. Please submit a PR so I can merge it here. Here you can find the [steps](https://github.com/allinurl/goaccess/issues/858#issuecomment-320257517) to do the translation if needed.

Let me know if you have any questions. Thanks!

PR #457, opened 2016-06-30
Add compile option about internationalization lables

charpty:
Add a compile option(--with-custom-lables) about internationalization lables.
Usage:   ./configure --with-custom-lables=lables_zh.h
This is helpful to show our own national language on the web page.

I made a simple demo example for this.
See here please: http://office.ifugle.net:20000/rap-logs/tmp/report.html  

Signed-off-by: Bo Cai charpty@gmail.com

PR #453, opened 2016-06-24
adding french translation

tconstans:
Hello,

Just created a quick (and probably) dirty french translation of ui.h file.

Best regards

PR #712, opened 2017-04-03
WIP add translation with gettext

haricot:
I think it's a good start.
But when I replace gettext_noop(Marks the given translation for extraction and returns it unchanged) "N_" to gettext "_"

Example in src/ui.h : 
#define VISIT_LABEL N_("Visitors") to #define VISIT_LABEL _("Visitors") 

This error appears when compiling make:
Error: An element of the initialization is not a constant

The solution may be to move the initialization in the main, in the initialization function.
I do not see too much! Do you have an idea ?

Ref issues PRs:
#701 #457 #453


allinurl:
This is a great start. Thank you!

I'll look into it and merge. Are you currently getting the error mentioned above with this PR? or do I need to change something to trigger it?

haricot:
I don't have the above error in the PR.
Yes ! We need to change a somethings.

Only the translations with `_()` in the src/ui.h files are translated into the program.
With `N_()` the translations are unchanged in the program but we can start translating them with .po files.

It would be necessary that all the translations begin with `_()` instead of `N_()`, it works with some translation because it does not use the same way in the .c

Try to find out why, but it does not come.

allinurl:
Thank a lot! I believe it should work now. Please give it a shot and let me know how it goes.

There are a few minor changes I need to do to the terminal dashboard so labels fit well, and probably bootstrap a JSON string into to the HTML document so labels in the [template file](https://github.com/allinurl/goaccess/blob/master/resources/tpls.html) get translated.

I need to check but I suspect there are still a few more string literals in the code that need to be added as `#defines` under ui.h.

haricot:
Very good !
Your changes resolve the issue.
Other translations appear and it's really nice! :-)
Indeed, there are still things to translate.
Thank you very much !

PR #717, opened 2017-04-05
Update fr.po

haricot:
Translation errors have occurred due to the change of order in the code, I suppose.
After a complete re-reading, I think it's good now.


allinurl:
Thanks. I'll be pushing some changes and additional translations soon.

Issue #727, opened 2017-04-07
Real-time html report not reloading using Docker


davask:
Hi,

here my goaccessconf file. 
I can t make it _real time html with automatic refresh_. What am I missing ?

fyi, I m using the docker version of allinurl/goaccess

```conf


time-format %H:%M:%
date-format %d/%b/%

log-format %h %^[%d:%t %^] "%r" %s %b "%R" "%u" %T %^
config-dialog true
hl-header true
html-report-title dwl web stats
json-pretty-print true
no-color false
no-column-names false
no-csv-summary false
no-progress false
no-tab-scroll false
with-mouse true
addr <my public ip>
origin http://<my public ip>
real-time-html true
ws-url <my public ip>:<my port>
log-file /dwl/var/log/apache2/access.log
debug-file /srv/data/debug.log
config-file /srv/data/goaccess.conf
invalid-requests /srv/data/invalid.log
no-global-config true
agent-list false
with-output-resolver false
http-method yes
http-protocol yes
output-format /srv/data/index.html
no-query-string false
no-term-resolver false
444-as-404 false
4xx-to-unique-count false
all-static-files false
double-decode false
ignore-crawlers false
crawlers-only false
ignore-panel REFERRER
ignore-panel KEYPHRASE
real-os true
static-file .css
static-file .js
static-file .jpg
static-file .png
static-file .gif
static-file .ico
static-file .jpeg
static-file .pdf
static-file .txt
static-file .csv
static-file .zip
static-file .mp3
static-file .mp4
static-file .mpeg
static-file .mpg
static-file .exe
static-file .swf
static-file .woff
static-file .woff2
static-file .xls
static-file .xlsx
static-file .doc
static-file .docx
static-file .ppt
static-file .pptx
static-file .iso
static-file .gz
static-file .rar
static-file .svg
static-file .bmp
static-file .tar
static-file .tgz
static-file .tiff
static-file .tif
static-file .ttf
static-file .fl
geoip-database /srv/data/GeoLiteCity.dat

```

allinurl:
I'd try commenting out `addr`and `origin` just to be sure those are not the conflict options. Also, please take a look at your browser's console and see if you can spot some errors on the client side.

davask:
I did it however I got this error in the console

```js

Uncaught DOMException: Failed to construct 'WebSocket': The port 65535 is not allowed.
    at Object.setWebSocket (http://163.172.35.42:65534/:3396:16)
[...]

PR #728, opened 2017-04-09
Expose default Goaccess port

julianxhokaxhiu:
Move from 8080 to 7890.

This should avoid confusion like in #727. Since no `goaccess.conf` is provided by default, use the default port where goaccess lives on.


allinurl:
Merged. Thank you!

PR #736, opened 2017-04-14
Update fr.po

haricot:
Still some small corrections on the chords and accents.
This should be right now.


allinurl:
Merged, Thanks!

PR #768, opened 2017-05-13
tc_db_set_path: accept db path with no trailing slash

terceiro:


allinurl:
Merged. Thank you for the PR!

PR #770, opened 2017-05-16
websocket: use uppercase method name only

### LINKED ISSUE #448: WebSocket error 400
I've got the following error when using web socket:

>WebSocket connection to 'ws://site.com:9879' failed: Unexpected response code: 400

I run go access with the following parameters:

```bash
goaccess -f /var/log/nginx/access.log --log-format COMMON -o /var/www/stats/index.html --real-time-html --port=9879 --ws-url=site.com --origin=site.com
```

mlosev:
fixes #448

`get` substring may happen in additional headers (i.e. `Host:` one)
so parsing will fail resulting in 400 return code (Bad request)

### DIFF
--- a/src/websocket.c
+++ b/src/websocket.c
@@ -1094,7 +1094,7 @@ ws_set_header_fields (char *line, WSHeaders * headers)
   if (line[0] == '\n' || line[0] == '\r')
     return 1;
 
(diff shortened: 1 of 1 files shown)


allinurl:
Merged. Thanks for reporting this and submitting this PR!

PR #774, opened 2017-05-17
Fix startup websocket race

xobs:
These two patches solved a problem I was experiencing when doing:

    ssh -i [key] core@remote.local docker logs -f [container] | ./goaccess -o report.html -a --log-format=COMBINED --real-time-html

The root cause is that the primary thread runs before mkfifo() is called.  Running goaccess under valgrind or "strace -ff" was usually enough to get it working, but this patch fixes this race condition.


allinurl:
Merged. Thank you!

PR #775, opened 2017-05-17
js: app: use wss:// when connecting via https

xobs:
This solves a problem I was experiencing on Firefox, where I had to manually specify --ws-url=wss://hostname

It is now inferred that the protocol should be wss:// when connecting via https.


allinurl:
Merged. Thanks for submitting this PR.

PR #896, opened 2017-09-24
Add log-size switch handling: [-S|--log-size]

TerraTech:
This is useful when piping in logs for processing in which the log size can be explicitly set.


TerraTech:
FWIW, I am using a little gzip helper that I wrote to feed in the gzip uncompressed filesize.
https://github.com/TerraTech/gzipInfo

allinurl:
Merged. Thanks for submitting this PR!

PR #904, opened 2017-09-28
goaccess.c: Curses should not be initialized if '--process-and-exit' …

TerraTech:
…is in effect

  This amends commit: dcc5bbdc


allinurl:
Merged. Thank you!

TerraTech:
You're welcome.  :)

Issue #907, opened 2017-09-30
Something amiss with bad exit code (1)


TerraTech:
I have run into a problem with goaccess returning a '1' exit code, when the processing completed without any errors that I could find.

My processing script snippet (simplified):
```bash
cat access.log | pv -l -F "${log} %b/$(getHuman ${cnt}) %r%a %p %t %e" -s "${cnt}" | goaccess --config-file="${f_config}" --db-path="${d_goaProcess}/" --process-and-exit --no-progress --no-parsing-spinner -f -
rc=$?
if (( rc )); then
    fatal "doit: goaccess failed: ${rc}"
    exit ${rc}
fi
```
I have added debugging instrumentation to the v1.2 branch.  
https://github.com/TerraTech/goaccess/commit/5f2aa58345ae7bd2b774895def5c0b9b00460187
_WITH_GETLINE_ is in effect

The output from my last run is:
```
[TT4-0] num_tests: 10
[TT4-1] test: 0, ret: 0
[TT4-0] num_tests: 10
[TT4-1] test: 1, ret: 0
TT3: 1
-
TT1: 1
+ rc=1
+ ((  rc  ))
+ fatal 'doit: goaccess failed: 1'
+ echo '[FATAL] doit: goaccess failed: 1'
[FATAL] doit: goaccess failed: 1
+ exit 1 
```
I haven't fully grok'd yet how 'test' is supposed to work, but from what I can tell it isn't being cleared properly in the flow of:  read_lines() => read_line(), where read_line() is supposed to 0 it out based on the result from pre_process_log().  Errantly, that 'test=1' is bubbling back up to main() and being used for the return.

@allinurl Can you please take a look and see if something obvious jumps out at you?  I'm thinking it is a simple fix, but I'm just not seeing it yet due to not fully understanding its scope.

TerraTech:
I may have found the problem and it only occurs when piping.  I added a quick little function print:
```c
static int
read_log (....)
{
...
fprintf(stderr, "read_log()\n");
```

Now the error output shows its being called twice for the single pipe:
```
read_log()
<<pv | goaccess processing output>>
[TT4-0] num_tests: 10
[TT4-1] test: 0, ret: 0
read_log()
[TT4-0] num_tests: 10
[TT4-1] test: 1, ret: 0
TT3: 1
-
TT1: 1
<snip>
```
That would explain why 'test' is never set to 0 and would then bubble up as main()'s return value.

TerraTech:
[...]

PR #908, opened 2017-09-30
options.c: Fix conf.filenames duplication problem if logs are via pipe.

TerraTech:
fixes allinurl/goaccess#907


allinurl:
Merged. Thank you!

PR #910, opened 2017-10-04
Add ability to store accumulated processing time into DB_GEN_STATS tcb file via '--accumulated-time' switch

TerraTech:
This was done to facilitate the processing flow of adding daily logs to the stored tcb files
    and then running the daily report against it.  This problem was that 'Init. Proc. Time' only
    showed the processing time for the current processed log, however we need it to show the
    accumulated time over the entire monthly report processing span.

### DIFF
--- a/src/goaccess.c
+++ b/src/goaccess.c
@@ -1487,6 +1487,17 @@ main (int argc, char **argv)
 
   /* ignore outputting, process only */
   if (conf.process_and_exit) {
+#ifdef TCB_BTREE
+    /* store accumulated processing time
+     * Note: As we store with time_t second resolution,
+     * if elapsed time == 0, we will bump it to 1.
+     */
+    if (conf.store_accumulated_time) {
+      time_t elapsed = end_proc - start_proc;
+      elapsed = (! elapsed) ? ! elapsed : elapsed;
+      ht_insert_genstats_accumulated_time (elapsed);
+      }
+#endif
   }
   /* stdout */
   else if (conf.output_stdout) {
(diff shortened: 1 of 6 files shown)


TerraTech:
@allinurl Please review this patch.  I wasn't quite sure where to wedge in the actual '--accumulated-time' switch in options.c as it is a bit of an oddball.  I also made this somewhat dependent on '#ifdef TCB_BTREE' where I felt it made sense to do so.

FWIW, I have processed 8 different monthly log sets amounting to over 50GB without any problems.  The resultant report shows the correct accumulated time values for each set.

TerraTech:
On another note, I have been backporting these changes to my FQ branch (v1.2).  Any chance of cutting a v1.3 or v2.0 release soon?

allinurl:
Merged and tested. Added a minor change as well. Thanks!

I'll certainly cut this release soon as it's way behind, it would probably be v1.2.1 as it contains mostly fixes. Thanks for the heads-up.

TerraTech:
I would say bumping to v1.3.x might be good for a couple reasons:
1. new switches are in play which changes/updates the user (UI/CLI) usage experience.
1. in semantic versioning (http://semver.org), M.m.p, where p==patch, are for bug fixes (re: <span>#</span>1 above)
1. and finally, my v1.2 backport branch is already up to: FQ/v1.2.10 (https://github.com/TerraTech/goaccess/releases) which will make my own versioning a slight bit awkward.
1. my company policy will only use official "stable" releases, however I have the flexibility to backport justified items from an upstream master branch, hence the FQ v1.2.x branch.

Just some things to consider.

PR #920, opened 2017-10-25
fgetline fgets() race handling

TerraTech:
This patch is to handle a fgets() + EOF/EAGAIN race in fgetline(), as well as adding two new switches to tweak its behavior.

I stumbled on this when I changed my log feeder to do something like the following:
```sh
{ printf "<custom crafted first log line>\n"; ssh remote gunzip -c logfile; } | pv | goaccess ....
```

The delay would sometimes cause the fgets() to go in a read loop and never exit in the face of either an EOF or EAGAIN.

The ```--getline-min-read``` specifies the minimum number of lines that must be read before it will quickly exit the read loop on an EOF/EAGAIN.  This covers the case where the logs being fed is a combination of 'fast' (printf) and 'slow' feeds where the slow feed incurs a remote ssh connection delay.  This covers how EOF/EAGAIN is handled as it should honor it after log lines have been fed for it, but be very conservative and wait for the log stream to start.  Non-blocking stream reads can get a bit tricky as the onus is on the coder to cover all the corner cases.  (I hope that makes sense).

As a belt-n-suspenders approach, it also has a ```--getline-timeout``` which will break out of the loop after specified number of seconds.  The default is 10, and has worked well for our use case.

Please give it a lookover and let me know what you think.  Thanks!


TerraTech:
rebased to master and ready to merge

allinurl:
Thanks. I haven't been able to merged this as I'm trying to keep up with some bugs/requests that I plan to squeeze in the next release. I still want to run some tests on my side before merging this commit.

PR #928, opened 2017-10-30
Added "Tiny Tiny RSS" to browsers.c file.

Nerdmind:
Hi. I've added "Tiny Tiny RSS" to the "Feeds" category in `browsers.c`. [Tiny Tiny RSS](https://tt-rss.org/) is a self-hosted and web-based RSS feed reader with a user-agent string like: `Tiny Tiny RSS/17.4 (http://tt-rss.org/)`. I hope, this is helpful and I can contribute to your project with this pull request.

## Additional:
I compiled this, but in the terminal ui interface, only `Tiny Tiny` instead of `Tiny Tiny RSS` is displayed within the "Browsers" section (see attachment). I don't know if this is a bug or expected behavior. Maybe the string is cut because of spaces between the words?

ctj69vqe

To test this, I've compiled with `{"TinyTinyRSS", "Feeds"}` (without spaces) and a test log which contains such a user-agent: `TinyTinyRSS/17.4 (http://tt-rss.org/)`. Then `TinyTinyRSS/17.4` is displayed.

### DIFF
--- a/src/browsers.c
+++ b/src/browsers.c
@@ -297,6 +297,7 @@ static const char *browsers[][2] = {
   {"newsbeuter", "Feeds"},
   {"Wrangler", "Feeds"},
   {"Fever", "Feeds"},
+  {"Tiny Tiny RSS", "Feeds"},
 
   {"Pingdom.com", "Uptime"},
   {"UptimeRobot", "Uptime"},


allinurl:
Thanks for submitting this PR. It has been merged. 

Great question. First the parser determines if the substring (e.g., `Chrome`) specified on the list is found within the given UA, (e.g., `AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.79 Safari/537.36`). If it is, then it attempts to extract not just whatever it was specified on the list (e.g., `Chrome`) but a bit more data from it, such as the UA version (`Chrome/61.0.3163.79`). However, to make this work for most cases, the parser needs to know when to stop extracting data from it. In this case, it looks for one of the following stoppers `;)  ` and extracts until there.

So you can see why this becomes a bit of an issue when the actual substring contains a white space such as `Tiny Tiny RSS/17.4`. I'd need to figure out a different algorithm that would work with the great majority of the UAs as in the first example and those containing spaces in between. You can take a look at where this is being parse (lines [368](https://github.com/allinurl/goaccess/blob/master/src/browsers.c#L368) and [393](https://github.com/allinurl/goaccess/blob/master/src/browsers.c#L393)). If you have any thoughts on this, please feel free to post them here.

allinurl:
Alright, I just pushed a fix to this issue. It should now display the whole string. Please feel free to build from [**development**](https://github.com/allinurl/goaccess#build-from-github-development) and let me know how it goes. Otherwise it will be pushed out in the upcoming release.

Nerdmind:
Hi. Thanks for your feedback and the explanation how this algorithm works. I don't understand the C code but I've compiled it from _development_ and can say that it seems to work!

qlviip_t

PR #955, opened 2017-11-28
parser.c: fix warning: comparison of array 'hour' equal to a null pointer is always false

TerraTech:
https://travis-ci.org/allinurl/goaccess/jobs/305616842#L2488-L2491

```
  src/parser.c:2356:7: warning: comparison of array 'hour' equal to a null pointer
        is always false [-Wtautological-pointer-compare]
    if (hour == '\0')
        ^~~~    ~~~~
  1 warning generated.
```


allinurl:
Merged. Thanks very much!

PR #956, opened 2017-11-28
settings.c: fix warning: expression which evaluates to zero treated as a null pointer constant of type 'char *'

TerraTech:
https://travis-ci.org/allinurl/goaccess/jobs/305616842#L2496-L2499

```
  src/settings.c:238:22: warning: expression which evaluates to zero treated as a
        null pointer constant of type 'char *' [-Wnon-literal-null-conversion]
    _argv[*argc + 1] = '\0';
                       ^~~~
  1 warning generated.
```


allinurl:
Merged. Thank you!

PR #957, opened 2017-11-30
Added 'Android 8 Oreo' to the list of OSs.

Arnie97:
The Android 8.0 'Oreo' update has been out for a while now.
Even Android 8.1 has released its final developer preview, and will release to the public in December.


allinurl:
Merged! Thank you sir!

PR #1036, opened 2018-03-02
Add SSL support for docker goaccess build

stoyandimov:
Addresses: [Goaccess docker ssl-cert unrecognized #1030](https://github.com/allinurl/goaccess/issues/1030)

I cloned the goaccess repo, changed the docker file as follows:
* apk added [openssl-dev](https://pkgs.alpinelinux.org/package/v3.3/main/x86/openssl-dev) similar to [libssl-dev](https://packages.debian.org/jessie/libssl-dev) (couldn't find libssl-dev for `apk`)
* added `--with-openssl` to the goaccess build configuration (in Dockerfile)

I built and tested the resulting image with valid certificates and a socket connection with `wss://` worked.

@allinurl  let me know what you think. If approved, I'll update the Readme file and docs as well.

PR #1038, opened 2018-03-05
Update README.md with wss support in Docker

stoyandimov:
Provides an example of configuring a secured connection
with `ssl-cert` and `ssl-key` config properties
in the Docker section of README.md.

(as promised in PR #1036)


allinurl:
Merged. Thank you!

PR #1040, opened 2018-03-06
Allow static file requests to be ignored

Jonuz:


Jonuz:
I changed the code so count_valid() can be skipped without having to ignore the request, thus making it possible to not raise valid requests count while still showing request in panels.

I'm not sure if the change in ignore_line() is too inconstant or if there is better way to achieve  it so I'm not merging it to this request.
https://github.com/Jonuz/goaccess/commit/1e2b0d87fecebc397bd817e5f3466ee129702dca

allinurl:
Merged. Thank you!

allinurl:
> I changed the code so count_valid() can be skipped without having to ignore the request, thus making it possible to not raise valid requests count while still showing request in panels.

> I'm not sure if the change in ignore_line() is too inconstant or if there is better way to achieve  it so I'm not merging it to this request.

Let me take a look at your changes and I'll make some updates if needed.  Thanks.

allinurl:
I made a few minor changes to your PR. I do like the changes you have in [Jonuz/goaccess@1e2b0d8](https://github.com/Jonuz/goaccess/commit/1e2b0d87fecebc397bd817e5f3466ee129702dca), the only thing I'd change is `--ignore-statics=1|2`, to be ` --ignore-statics=req|panel`, that way we can even use the same style across all `--ignore-*` options.

Please feel free to submit another PR and I'll be happy to merge it. Thanks again.

ristein:
I like this option but would like to also skip "not found urls" from being count as well as those "static requests".
Also it would be nice to have the "static requests" field not empty

PR #1041, opened 2018-03-06
Allow static file request to be not counted as valid without ignoring it

Jonuz:
* Added IGNORE_LEVEL_PANEL and IGNORE_LEVEL_REQ definitions
* Allow request to be not counted as valid without ignoring it.


allinurl:
Merged. Thank you!

PR #1064, opened 2018-03-30
uniq (default) dbpath for each process

gitqlt:
When we start multiple (btree enabled) goaccess processes, each of them gets a uniq folder for its database files. (Only if no db-path is specified, neither in the config file nor with the --db-path CLI argument). This makes parallel processing possible in an easy way. The extra folder is removed at the housekeeping.


allinurl:
Merged. Thank you!

Issue #1068, opened 2018-04-03
Build parameters


szepeviktor:
```
$ goaccess --version
GoAccess - 1.2.
For more details visit: http://goaccess.io
Copyright (C) 2009-2016 by Gerardo Orellana
```
Many C programs display the `./configure` parameters in version info.
Please consider adding [info](https://github.com/allinurl/goaccess/blob/master/configure.ac#L286-L297) (openssl, debug, utf8, geoip, storage, compression) to `--version` output.

allinurl:
For some reason I thought somebody asked about this before, but I couldn't find it. I can certainly do that. Thanks!

allinurl:
Implemented #1101.

szepeviktor:
Thanks

Issue #940, opened 2017-11-09
Implement a command line argument to show the configure options at compile time


ms217:
As subject is suggesting it.
Afaik GoAccess doesn't support this. At least I haven't found this option in the latest stable 1.2 release.

An example with NGINX: I can check the configure arguments when I issue "nginx -V".
I need this for a GoAccess implementation into a webadministration panel.

millsjustin:
Duplicate of #1068.
Implemented in #1101.

allinurl:
Closing this. Thanks!

PR #1101, opened 2018-04-28
added configure arguments to --version output

millsjustin:
I added configure arguments to the `--version` output. As suggested in #940 

If anything is wrong or you would like this implemented in different way, I would be happy to change anything. 😃

I thought this had been referenced in another issue also but I can't find it.


allinurl:
Merged. Thanks a lot!

allinurl:
This closes #1068.

Issue #902, opened 2017-09-28
UI: goto beginning, goto end


TerraTech:
On the report.html page, in addition to the goto prev (<) and goto next (>), it would be nice if there were goto beginning (<<) and goto end (>>).  One one report, the "VISITOR HOSTNAMES AND IPS" has 366 line items.

e.g.: **<<** **<** **>** **>>**

I've seen some UI's use:  **<** **>** **<<** **>>**

millsjustin:
Implemented in #1102.

allinurl:
Closing this. Thanks!

PR #1102, opened 2018-04-29
added first/last buttons to html report

millsjustin:
I added `First` and `Last` buttons to the html report. As suggested in #902 

This isn't ready to merge yet because it looks like you've created a min version of Font Awesome that only has the icons needed for the report. How did you make this? Right now the `First` and `Last` buttons are using the same icons as `Previous` and `Next`. Is there a way to add icons for `<<` and `>>`?

Also the double chevrons to match the chevrons for `Previous` and `Next` are part of the pro Font Awesome. So it might be better to use `fa-angle-left` and `fa-angle-double-left` so they match.

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -708,14 +708,42 @@ GoAccess.Panels = {
 			$pagination.parentNode.classList.add('disabled');
 	},
 
+	enableFirst: function (panel) {
+		var $pagination = $('#panel-' + panel + ' .pagination a.panel-first');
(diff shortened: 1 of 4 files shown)


allinurl:
Merged. This is awesome :) thank you so much!

millsjustin:
I think this would look better if the `First` and `Last` buttons didn't use the same icon as `Previous` and `Next`, but I wasn't sure how to add more icons.

Issue #1056, opened 2018-03-27
HTML time zone


szepeviktor:
It would be nice

- to have an option for setting the time zone for the HTML output
- to have another option to auto-adjust the time for the browser's time zone `(new Date()).getTimezoneOffset()`

allinurl:
Would this for the `last updated` date and/or for the Visitors panel? similar to #496?

szepeviktor:
both

r7l:
Is there any progress for this? Still having the issue that everything is +0000.

allinurl:
@r7l not yet. If it's a change on the HTML report only, then it should be fairy trivial to do it in JS.

Issue #496, opened 2016-08-18
add support to timezone


flyisland:
right now, we have to use "%^" to ignore the timezone like "+0800". but it'll be confused while the web server used UTC format for the request time.

allinurl:
Thanks for suggesting this. Can you please elaborate a bit more how the timezone should be displayed in goaccess and the use case you have in mind?

flyisland:
my suggestion:
1. goaccess uses UTC internally
2. provide `timezone` setting in UI

allinurl:
I'm not sure I follow. Are you looking to convert the given log time to a specific timezone? which panel would be affected with this change? Right now goaccess simply uses/displays the date and time that the access log has on it.

flyisland:
"08/Aug/2016:16:49:16 +0800" and "08/Aug/2016:08:49:16 +0000" actually are the same time, but will be different in goaccess since goaccess will simply ignore the `timezone` field.

I sometimes have to use goaccess to analyze the performance of websites which with different datetime format setting, ignoring the `timezone` will be confused in some situation.

allinurl:
Got it. I'll look into this. Thanks for the reply!

ray-moncada:
How do I set the timezone when creating an HTML report?

allinurl:
@ray-moncada are you asking about the last updated time zone? or for each metric? If the latter, then it uses the tz from your access log.

iskandarsaleh:
hi, @allinurl 
would you mind that i have example log like this : 
2020/07/20 14:53:27.396 [I] [inquiry.go:47]  Request Inquiry QRIS:  {"dataQr":"https://support.lenovo.com/qrcode?sn=PF11UNLP&mtm=80XK0052ID","issuer":"JATxxx","custAccount":"0812xxx"}

how can i setting this kind of log to view on goaccess 

for now i get error : goaccess: unrecognized option '--3' 

thanks before


regards

allinurl:
@iskandarsaleh seems like there's no valid IP. goaccess requires a valid IPv4/6, a request and date/time.

alaskapublic:
> @allinurl This is my scenario. The server time is +0000, my local is time is -0500. When goaccess static report is generated, is there a way to have the report show my local time zone?

I'll second this issue. 

Server time is in UTC, and local time is -8 or -9 depending on Daylight Savings Time.
Problem is that the logs are all shifted by 8 or 9 hours. 

This makes it problematic when looking at:
[...]

PR #1104, opened 2018-04-30
Parsing Time Zones

millsjustin:
I've been looking at #1056 and #496

As long as the date and time can be parsed as the same field in the log line it seems that these could both be implemented with the `%x` specifier. `strptime` and `strftime` will handle the timezone conversions if you specify the timezone offset with `%z`. The output times will be converted to the local timezone defined by the `TZ` environment variable.

Example Log Lines:
```
1.1.1.1 - - [29/Apr/2018:22:00:00 -0200] "GET / HTTP/1.1" 200 185
1.1.1.1 - - [30/Apr/2018:00:00:00 +0000] "GET / HTTP/1.1" 200 185
1.1.1.1 - - [30/Apr/2018:02:00:00 +0200] "GET / HTTP/1.1" 200 185
```

Formats:
```
log-format %h %^[%x] "%r" %s %b
time-format %d/%b/%Y:%H:%M:%S %z
date-format %Y%m%d
```

When I run this log with these formats I get three hits all for the same time. The only problem was that when I specified `date-spec hr` things broke when the time was appended to the date. So I add the `set_time` method to format the `logitem->time` field, similar to what `set_date` does for the `logitem->date` field.

It does seem kind of strange to have all the date/time format defined in the `time-format` though and the `date-format` needed to have this format string for this to work.

So I'm not sure how much this really helps but I just wanted to share what I found.

Also I haven't tested how this will impact parsing performance.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -964,6 +964,23 @@ set_date (char **fdate, struct tm tm)
   return 0;
 }
 
+/* Format the broken-down time tm to a numeric time format.
+ *
+ * On error, or unable to format the given tm, 1 is returned.
+ * On success, a malloc'd format is returned. */
+static int
+set_time (char **ftime, struct tm tm)
+{
+  char buf[TIME_LEN] = "";
+
+  memset (buf, 0, sizeof (buf));
+  if (strftime (buf, TIME_LEN, "%H:%M:%S", &tm) <= 0)
+    return 1;
+  *ftime = xstrdup (buf);
+
+  return 0;
+}
+
 /* Determine the parsing specifier error and construct a message out
  * of it.
  *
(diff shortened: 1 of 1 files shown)


allinurl:
Thank you! I merged this. I'm looking at what changes need to be done and I'll post back.

allinurl:
I was looking at this and you are right, I think going forward we'll need to introduce another command line option such as `--date-time-format=%d/%b/%Y %H:%M:%S %z` in favor of  `--date-format` and `--time-format`.

millsjustin:
I had the idea to always combine the `date` and `time` fields and parse them at the same time, but that seemed to slow down parsing quite a bit because of the extra calls to `malloc` and `memcpy`. You can see what I did here: https://github.com/allinurl/goaccess/compare/master...millsjustin:parse-datetime

I think you're right that adding a `--date-time-format` option would be best. I've started working on that and almost have it working. I'll follow up when it's done.

PR #1110, opened 2018-05-04
added missing call to free()

millsjustin:
Sorry I missed this. After the call to `set_time` the token can be freed.


allinurl:
Merged. Thank you.

PR #1131, opened 2018-05-29
Make regular expressions POSIX compatible

millsjustin:
This is a follow up to the comments on 4b8346e

The default `sed` on mac does not support the `*?` non-greedy operator. Which is why `resources/js/charts.js.tmp` and `resources/js/app.js.tmp` were empty and I was getting the error: `RE error: repetition-operator operand invalid` So I removed the non-greedy operator. I don't think it's needed in this context.

I also noticed that the files produced by the default sed on mac were not the same as files produced by GNU sed. I replaced `[ \t]` with `[[:space:]]` and now files produced by default mac sed and GNU sed are identical.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -46,7 +46,7 @@ CLEANFILES =                          \
 # Tpls
 tpls.h: bin2c$(EXEEXT) resources/tpls.html
 if HAS_SEDTR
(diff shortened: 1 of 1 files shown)


allinurl:
Thanks a lot for fixing this!

PR #1153, opened 2018-06-19
Unset referrer header

da2x:
Server statistics is assumed to be a private page, so don’t send the HTTP Referer (sic) request header with any external links to keep information about the existence of the page from leaking.
https://www.w3.org/TR/referrer-policy/#referrer-policy-no-referrer
https://www.ctrl.blog/entry/private-bts-referrer-header


allinurl:
Merged. Thank you!

PR #1169, opened 2018-07-03
Speed up animations

pravdomil:
if you click to expand left panel, it takes too long to animation completes


allinurl:
Merged. Thank you!

PR #1200, opened 2018-08-14
Rewrite extract_method

steelywing:
A loop feel like better, if you don't mind it calculate the string length on fly, please remove the below part (will be cleaner), and use `strlen()` instead of `list_length`.
```c
  // Length of every string in list
  static int list_length[sizeof(list) / sizeof(*list)] = { -1 };
  // Only calculate length on first time
  if (list_length[0] == -1) {
    for (int i=0; i < list_count; i++) {
      list_length[i] = strlen(list[i]);
    }
  }
```

PS. I am not good in C, I have tested but please test again 😊 .

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -716,66 +716,38 @@ verify_static_content (const char *req)
 static const char *
 extract_method (const char *token)
 {
-  const char *lookfor = NULL;
+  const char *list[] = {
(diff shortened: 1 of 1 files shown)


allinurl:
Merged. Thank you for submitting this PR. I tested this out, went well — I'll benchmark this and see if it's worth moving the  `list_length` loop outside of `extract_method()`.

PR #1201, opened 2018-08-15
Simplify extract_referer_site

steelywing:
We don't need to change the `url`, so just use the `referer` is fine? (actually I just want no `goto`)


allinurl:
I'm closing this as I think the issues brought up have been fixed. Otherwise, feel free to reopen it. Thanks!

PR #1285, opened 2018-10-26
Add Japanese translation

4f8p:
Japanese lang file. 🗾
(Since there was a mistranslation, I've fixed it.)

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -1,3 +1,4 @@
 fr
 es
 zh_CN
+ja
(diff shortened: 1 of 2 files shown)


allinurl:
Merged. Thank you for submitting for this PR - Dōmo arigatō!

4f8p:
Anytime! 👍
I don't know whether it can be translated correctly, but we will send you a revised version if there is a mistranslation.

PR #1293, opened 2018-11-05
Added some crawlers

EifX:
Hi, I have some crawlers found in my log-files. I don't know what PxBroker really is, but Seekport is an old search engine (currently disabled) but their bots seems to be active...

Greetings,
Alex


allinurl:
Merged. Thank you!

PR #1296, opened 2018-11-06
Added somme crawlers

chmoelders:
Hi!
I would like to add some crawlers as well. The named crawlers are mostly active in germany.

Regards,
Christian


allinurl:
Merged. Thank you!

PR #1347, opened 2019-01-08
Typo config file

carlosduelo:
typo fifo-out instead of fifo-in


allinurl:
Awesome, thanks for reporting this. I completely missed.

PR #1351, opened 2019-01-11
Fix fallback to global config

sthen:
#694 results in ~/.goaccessrc always being used, the fallback to the global config file is broken. It appears that the code is assuming that realpath() will return NULL if the file does not exist, which is not the case. Unbreak by actually attempting to open ~/.goaccessrc and falling back to global config if this fails.


allinurl:
Thanks a lot for this PR. It has been merged.

PR #1353, opened 2019-01-15
Fix parse_raw_data for VISITORS module when using Tokyo Cabinet

martinsp:
Thanks for the great software.

I noticed that when using Tokyo Cabinet btree storage the VISITORS section of HTML report randomly changed dates. This patch fixes the issue.

The issue is shown in this youtube video: https://youtu.be/yp8jgVLTWs8


allinurl:
Merged. Thank you very much!

PR #1370, opened 2019-02-04
Enable out-of-tree builds

vincentbernat:
autoconf/automake usually allows a user to do an out-of-tree build:

    mkdir build
    cd build
    ../configure
    make

This is convenient to easily start from scratch by just removing the
`build` directory (whatever we are working from a git checkout or a
tarball).

To make it work, manually written rules, like these used to embed
HTML/JS/CSS resources, need to be rewritten to correctly address
source files.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -44,68 +44,76 @@ CLEANFILES =                          \
   resources/js/app.js.tmp
 
 # Tpls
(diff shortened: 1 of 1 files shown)


allinurl:
Awesome, didn't know about the out-of-tree build. Merged. Thank you!

PR #1372, opened 2019-02-04
Add an option to disable client IP validation

### LINKED ISSUE #1354: Compatibility with IP obfuscation
One way to ensure we don't collect personal data in logs is to obfuscate the IP address. Some users may just remove the last byte of the IP, but this makes it more difficult to count visits. On the other hand, for IPv6, this is not enough as the first 8 bytes are considered enough to identify most users.

nginx has a third-party module called [ipscrub](https://github.com/masonicboom/ipscrub). It hashes IP addresses to strings which do not look like an IP address. However, goaccess `%h` is not compatible with such an identifier.

Should we make `%h` more lenient? Or provide an alternative accepting this identifier?

Example:

```
PSe5mXpHEZ2zhB9OL+crFY - - [18/Jan/2019:06:31:27 +0000] "GET /fr/blog/atom.xml HTTP/1.1" 304 0 "-" "rss2email/3.9 (https://github.com/wking/rss2email)"
wdBIcM - - [18/Jan/2019:06:31:54 +0000] "GET /fr/blog/atom.xml HTTP/1.1" 200 44154 "-" "UniversalFeedParser/5.1.3 +https://code.google.com/p/feedparser/"
```

Long ones are IPv6, short ones are IPv4.

vincentbernat:
To avoid logging IP addresses, users may obfuscate them using a
regularly rotated seed and an hash algorithm. For nginx, [ipscrub][]
can be used for such a task. However, the result is not a valid IP
address.

This commit adds an option, `--no-ip-validation`, to avoid checking
the validity of an IP address and use the provided field as is. The IP
is stored as a string and the current code correctly handles the case
where this field is not an IP address (`inet_pton()` is checked with
`AF_INET`, then `AF_INET6` and falls back to some sensible behaviour
for other cases).

Fix #1354

[ipscrub]: https://github.com/masonicboom/ipscrub

### DIFF
--- a/config/goaccess.conf
+++ b/config/goaccess.conf
@@ -520,6 +520,11 @@ ignore-panel KEYPHRASES
 #ignore-status 400
 #ignore-status 502
 
+# Disable client IP validation. Useful if IP addresses have been
+# obfuscated before being logged.
+#
+# no-ip-validation true
+
 # Number of lines from the access log to test against the provided
 # log/date/time format. By default, the parser is set to test 10
 # lines. If set to 0, the parser won't test  any  lines and will parse
--- a/goaccess.1
+++ b/goaccess.1
@@ -608,6 +608,10 @@ ww?.domain.*
 Ignore parsing and displaying one or multiple status code(s). For multiple
 status codes, use this option multiple times.
 .TP
+\fB\-\-no-ip-validation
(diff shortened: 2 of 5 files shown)


allinurl:
It has been merged. Thank you for submitting this PR!

PR #1385, opened 2019-02-22
typo fix

kAworu:
Before this patch, a parsing error would look like:
> Parsed 1 linesproducing the following errors:


allinurl:
Merged. Thanks!

PR #1410, opened 2019-03-24
small fix

err931:
- Modify I name
- Fix duplicate contributor name
- Remove Vagrantfile (Not used)


err931:
Oops! Reopend.

allinurl:
Can I ask why removing the vagrant file? Thanks

err931:
**Of course, there is a reason for requesting deletion.**
The `Vagrantfile` has not been updated for more than three years and the base OS is an older version.

Problems that don't work may also be caused by bugs (or configuration errors) that occur from the **VirtualBox** side.
I decided to remove it because it could increase the running cost.

allinurl:
Got it. Thanks!

PR #1420, opened 2019-04-05
Fixes #1419: ignore-referer-report wasn't released under that name

### LINKED ISSUE #1419: Confusing references to both hide-referer and ignore-referer-report in changelog
https://github.com/allinurl/goaccess/blob/master/ChangeLog currently has the following entries for GoAccess 1.3:

>   - Added --hide-referer command line option to hide referers from report.
>   - Added --ignore-referer-report command line option to hide referers from
>     output.

This had me confused. Turns out, both entries mean the same feature which was checked in as `ignore-referer-report` initially and renamed into `hide-referer` later in the development cycle. Maybe remove that `ignore-referer-report` line so nobody else has to research this?

palant:
This feature has been released as hide-referer and is already documented in another line.


allinurl:
Merged. Thank you for sending this PR!

PR #1437, opened 2019-04-26
Add base64 encoded favicon to html output

### LINKED ISSUE #1432: HTML report /favicon.ico 404
You may notice that every time you open up the html report generated by the goaccess in the browser, your browser will request a favicon.ico file from the root uri. 
The [demo site](https://rt.goaccess.io/?20181123091935) also have this issue.  
When the `image/x-icon` is not set in the html file, the browser will try to request a favicon.ico file from "https://rt.goaccess.io/favicon.ico" which may cause a 404.And it takes up time. 
Many of us won't put the favicon.ico in "/", instead it will be put somewhere else, the path will be set by the html file like
```html
<link rel="icon" href="somewhere/favicon.ico" type="image/x-icon" /> 
```
for goaccess maybe a better solution is to base64 the favicon.ico file and embed it into the html report file, the code may looks like this:
```html
<link href="data:image/x-icon;base64,AAABAAEAEBAQAAEABAAoAQAAFgAAACgAAAAQAAAAIAAAAAEABAAAAAAAgAAAAAAAAAAAAAAAEAAA
AAAAAADGxsYAWFhYABwcHABfAP8A/9dfAADXrwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAIiIiIiIiIiIjMlUkQgAiIiIiIiIiIiIiIzJVJEIAAAIiIiIiIiIiIiMyVSRC
AAIiIiIiIiIiIiIRERERERERERERERERERERIiIiIiIiIiIgACVVUiIiIiIiIiIiIiIiIAAlVVIi
IiIiIiIiIiIiIhEREREREREREREREREREREAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" rel="icon" type="image/x-icon" />
```

dragonandy:
If html file didn't have a icon link,browser may request a facicon.ico
from the root uri which may not have on the server, and it will return
a 404 error code.
this might not work for browsers lower than IE8

Fix #1432

I am not familiar with c ,this is the best i can do :)

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -232,6 +232,16 @@ print_html_header (FILE * fp)
   "<meta name='viewport' content='width=device-width, initial-scale=1'>"
   "<meta name='robots' content='noindex, nofollow'>", _(DOC_LANG));
 
+  /* Output base64 encoded goaccess favicon.ico*/
+  fprintf (fp, "<link rel='icon' href='data:image/x-icon;base64,AAABAAEA"
+  "EBAQAAEABAAoAQAAFgAAACgAAAAQAAAAIAAAAAEABAAAAAAAgAAAAAAAAAAAAAAAEAAAA"
+  "AAAAADGxsYAWFhYABwcHABfAP8A/9dfAADXrwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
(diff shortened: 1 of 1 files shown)


allinurl:
Merged. Thank you!

Issue #1404, opened 2019-03-13
Proposal: Renew Dockerfile


err931:
I have used the **GoAccess** in business. It has achieved great results and is well received in the company.

However, the size of official Docker image is so large that security audits may not pass.
So I wrote an unofficial Dockerfile and operated it in-house. It has been in operation for about half a year, but it's stable enough.

I will submit `Dockerfile` if you accept it.
Apply **Apache License 2.0** to the license. This does not conflict with the MIT license.

Please kindly consider this matter. :muscle:

# Comparison
- Using a multi-build, we reduced the size of the image.
	- After building in Alpine, it will be saved in the image of BusyBox. By eliminating extra packages, you are more resistant to vulnerabilities.
- Replace OpenSSL to LibreSSL.
	- Normally, you can't use host-side shared libraries in container image. Therefore, I replaced it with a static link.
	- Less susceptible to vulnerabilities from OpenSSL, you can avoid unnecessary build.
- Compatible with `ARM architecture`.
	- This image has been successfully built many times on `Raspberry Pi 3`.
	- So you don't have to worry about `ARM-based CPUs` that will increase in the future.
- Minimum required command.
	- There is no need to insert `tini` to make it **PID 1**. It can be substituted with the `docker run --init` command.
	- Even using the `VOLUME` command, you can't host side and the synchronization.
	- Only `/var/www/goaccess` and 7890 port are open to connect with `nginx` container etc.

allinurl:
Would it work still the same way as it's in the [**docs**](https://github.com/allinurl/goaccess#docker)? I'm all for any improvements and this sounds like it has several of them :)

err931:
Regarding your question, **most of them work.** :+1:
However, I have disabled **GeoIP** and **Tokyo Cabinet**.

Not to mention the latter, but there is a reason for **GeoIP**. Because it must respond to [GDPR](https://ec.europa.eu/info/law/law-topic/data-protection_en).
**GDPR also protects IP addresses**, which must be excluded if a viewer refuses to retrieve data.

However, it's difficult to record that intention in the access log.
The only workaround is to `Anonymize all IP addresses` in which case GeoIP won't work.

Can you tell me what you think is a problem? Answer your questions.

# For example
If you want to output Apache log as HTML report, you can do as follows. There is no need to install directly as it's isolated in the container.
**Note: if you were replaced by the new Dockerfile.**

```
[...]

PR #1453, opened 2019-05-12
rewrite Dockerfile

err931:
:warning: **This request makes destructive changes. Please verify and merge.**

Links: #1404

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,63 @@
-# Builds a goaccess image from the current working directory:
-FROM alpine:edge
+# Copyright 2019 Hiroki Kamino.
+# Originally under MIT in Gerardo Orellana. https://goaccess.io/
+#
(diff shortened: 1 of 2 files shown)


allinurl:
Merged. Thanks very much. 

BTW, can I ask what destructive changes it does?

err931:
> BTW, can I ask what destructive changes it does?

**I wrote to the issue, but has changed the library from OpenSSL to LibreSSL.**
There are no compatibility issues with OpenSSL, but there may be new issues that were not noticed during testing.

I think so...

frebib:
> BTW, can I ask what destructive changes it does?
Undocumented behaviour change when the user doesn't override the CMD. This just bit me :(

allinurl:
@err931 Could you please address @frebib question? Thanks!

PR #1516, opened 2019-07-30
Typo in goaccess.conf

chang-zhao:
I understand that `crawlers-only` is the opposite part to `ignore-crawlers`, so the comment can't be `This will ignore robots listed under browsers.c`?


allinurl:
Merged. Thank you!

Issue #1520, opened 2019-08-07
minor typo in README.md


dmaziuk:
I don;t htink this is worth a separate pull request:

`cat access.log | docker run --rm -i -e LANG=$LANG allinurl/goaccess -a -o html --log-format COMBINED > report.html`

is missing a dash for input file name. I.e. the above as is prints usage message, with `... COMBINED - > report.html` it works.

PR #1521, opened 2019-08-08
Typo: Docker commandline example doesn't work

markiewb:
Fixes https://github.com/allinurl/goaccess/issues/1520


allinurl:
Merged. Thanks for spotting that!

PR #1522, opened 2019-08-08
Add info about predefined log-format options

markiewb:
This was a showstopper for me, when I tried the docker variant to analyse a Tomcats access log.


allinurl:
Merged. Thank you!

PR #1529, opened 2019-08-14
Fix time-zone is not reflected

err931:
# Changes
- Get time-zone setting from environment variable
- Also modified `README.md`
- Remove unused packages and sorted


allinurl:
Merged. Thank you!

PR #1578, opened 2019-10-25
added list with podcast-related agent strings

rahra:
This is a list of podcast related agent strings.
It may be useful for people who host their podcast media files on their own servers.


allinurl:
Merged. Thanks a lot for submitting this PR!

PR #1608, opened 2019-11-20
Add missing gettext build requirement

### LINKED ISSUE #1547: Additonal dependency to build on RHEL/CentOS
When building 1.30 it looked like I needed to add gettext-devel to the requirements on CentOS 7. I was going to PR the readme but was not sure if the additional dependency is correct or not. I figured I would ask first just in case.

The spec file I used as reference is the following:
https://src.fedoraproject.org/rpms/goaccess/blob/master/f/goaccess.spec

evitalis:
I was not sure what was required for Gentoo or Slackware as I don't have access to those platforms so I left it blank for the moment. Did not see a development branch to issue this PR to in the interim but feel free to update if needed.

Resolves #1547


evitalis:
This was totally broken and trying to rebase led to a giant headache. I am closing this and resubmitting a clean PR.

PR #1613, opened 2019-11-26
List of podcast related agents updated

rahra:
Minor update of the list, some new podcast-related agents.


allinurl:
Merged, thanks for submitting this!

PR #1659, opened 2020-01-22
Add missing gettext build requirement

### LINKED ISSUE #1547: Additonal dependency to build on RHEL/CentOS

evitalis:
Resubmitting this as the previous PR was broken.

Resolves #1547

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,13 +187,13 @@ GoAccess has minimal requirements, it's written in C and requires only ncurses.
 However, below is a table of some optional dependencies in some distros to
 build GoAccess from source.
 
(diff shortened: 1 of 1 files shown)


allinurl:
Merged. Thank you!

PR #1661, opened 2020-01-30
Ignore SIGPIPE with SIG_IGN

mtorromeo:
Change the handling of SIGPIPE signals.

With the previous handler killing goaccess with SIGTERM makes it unresponsive, using 100% of the cpu and fails to terminate. By tracing the process I noticed a flood of SIGPIPE signals.

With this change the process now behaves correctly but I'm not really sure why the previous version manifested that behavior.


allinurl:
Thanks for submitting this. Can I ask you just so I can test out the issue on my end to share the steps on how to reproduce the bug? Thanks

mtorromeo:
Well, my setup is a bit convoluted to reproduce but I'm sure there is a simpler way to test it.
I'm spawning goaccess from another process with stdin and stdout attached to it (using the cockpit.spawn() api from cockpit-project.org).

At some point I want to terminate it and, from what I understand, the api closes stdin and stdout and then sends sigterm to the process. At this point the goaccess process hangs and never quits.
Inpecting the process shows the cpu usage and sigpipe flood that I mentioned before.

allinurl:
Merged. Thank you. It will be pushed out in the upcoming release.

PR #1713, opened 2020-03-24
Add option user-name=username

rsmarples:
This makes it trivial to integrate into any OS's init system.

For example, it's not possible to launch from NetBSD's rc system as the www user because it's locked down so the su command does not work and a third party tool would be required.

It also allows pkgsrc to supply an init script so it can be daemonised easily.


rsmarples:
Arggg, there is an issue with the db files owner.... Will redo

PR #1718, opened 2020-03-27
Add option user-name=username

rsmarples:
This makes it trivial to integrate into any OS's init system.
GoAccess should not run as root when producing real time reporting.


allinurl:
This is great. Thanks for submitting this! Merged.

Issue #1086, opened 2018-04-16
add ssl_protocol distribution panel


VinceRafale:
Hello,

As an enhancement is it possible to have a panel displaying the ratio/usage of ssl_protocol?
This information is available by default for aws cloudfront logs for example.

Thanks

allinurl:
Sounds like we could add a panel dedicated for this. Do you have some sample data that I can look? Thanks.

MrSorcus:
Nginx log format
```
log_format goaccess "[$time_local] $remote_addr $http_host $ssl_cipher $ssl_protocol"
  "\"$http_user_agent\" \"$http_referer\" "
  "\"$request\" $status $bytes_sent $request_time";
```
Logs
```
[19/May/2019:16:47:55 +0000] 2a01:7e01::1 mail.example.com TLS_AES_256_GCM_SHA384 TLSv1.3 "Mozilla/5.0 (X11; Linux x86_64; rv:66.0) Gecko/20100101 Firefox/66.0" "https://mail.example.com/?_task=mail&_mbox=INBOX" "POST /?_task=mail&_action=refresh HTTP/2.0" 200 448 1.042
[19/May/2019:16:47:57 +0000] 2a01:7e01::1 example.com TLS_AES_256_GCM_SHA384 TLSv1.3 "curl/7.64.1" "-" "GET / HTTP/2.0" 403 234 0.058
```
Nginx version 1.15.12

MrSorcus:
Something similar for AWS here - https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html (`ssl-protocol` & `ssl-cipher`)

allinurl:
So essentially you want to display a panel containing the TLS version and cipher?

MrSorcus:
Not sure how must it be present in GoAccess.

whalehub:
@allinurl I'm also interested in this feature. My use case is monitoring the rate of support for TLSv1.3 among the visitors of my site so I can make informed decisions in regards to deprecating the use of earlier versions.

Here's a quick mock-up of what I imagine the corresponding panel to look like:

![vivaldi_2020-06-09_16-46-18](https://user-images.githubusercontent.com/44198148/84163018-5312f880-aa71-11ea-9cff-6a962381c336.png)

allinurl:
[...]

PR #1727, opened 2020-04-02
Logging MIME-Type and TLS parameters

uschwarz:
This adds support for analyzing MIME-Type (such as `text/plain`, `%{Content-Type}o` in Apache) and TLS connection parameters (such as `TLSv1.2 ECDHE-RSA-AES128-GCM-SHA256`, `%{SSL_PROTOCOL}x %{SSL_CIPHER}x` in Apache, but really it's just two strings separated by a space if people want key length or anything else in the second field).

You might want to give this a critical look before merging, C is not my native language by a long shot, but valgrind said I was fine in terms of memory leaks. (Also, I didn't much test the interactive mode since I mainly use goaccess for HTML reports.)


allinurl:
Thanks for posting this PR. Can I ask you, do you have some sample lines from an access log that contain these fields so I can test it out on my end? Thanks!

uschwarz:
Sure. These are 800 random somewhat anonymized lines, which my local build could parse with `--log-format='%h %^[%d:%t %^] "%r" %s %b "%R" "%u" %^ %^ %D %M %K'`

(Our Apache httpd logs `%I %O %D %{Content-Type}o %{SSL_PROTOCOL}x %{SSL_CIPHER}x` after the User-Agent.)

[sample.anon2.log](https://github.com/allinurl/goaccess/files/5541905/sample.anon2.log)

allinurl:
Merged. I tested out and works well. I did change slightly the extract routine so it can parse it regardless the order of `SSL_PROTOCOL` and `SSL_CIPHER`. It will be out in v1.4.2 this week.

Thanks again!

Closes #1086.

PR #1790, opened 2020-05-26
Add .dmg, .xz, and .zst to the static list.

geraldcombs:
.dmg is the most commonly used extention for Apple Disk Image files,
.xz is used by XZ Utils, and .zst is used by Zstandard.


allinurl:
Awesome, thanks for submitting this PR :+1:

PR #1802, opened 2020-06-06
Update German translation

cgzones:
2 notes:

* the help window strings are not translatable help window

* The texts in the footer might overlap, e.g. the translation of `[?] Help [Enter] Exp. Panel` on the left and the translation of `[q]uit GoAccess` on the right
  footer left
  footer right


allinurl:
vielen Dank!

PR #1817, opened 2020-06-19
README: update openSUSE instructions

moio:
The `zypper ar` line resulted in an error as a repo name is required.
Given the `-f` (autorefresh) option is specified, manually refreshing (`zypper ref`) is not really needed.


allinurl:
Merged. Thank you!

PR #1847, opened 2020-07-15
Use https for the website url

elboulangero:
For you to consider


elboulangero:
Note that you could also use the static string `PACKAGE_URL` which is already set to `https://goaccess.io` per https://github.com/allinurl/goaccess/blob/master/configure.ac#L5, and then completely get rid of the `GO_WEBSITE` definition.

allinurl:
Merged. Thank you for submitting this PR! I agree, `PACKAGE_URL` it's probably the way to go.

PR #1868, opened 2020-08-04
Log filepath on error

bagage:
I'm trying to use the dockerized version of goaccess but currently it fails with `Unable to open the specified log file. No such file or directory`. Since I'm not 100% sure about which log file it considers, having it in the error message can help!


allinurl:
This has been implemented.

Issue #1693, opened 2020-03-04
How to analyze zipped log files on a remote webserver?


JoshDebris:
I want to set up goaccess on my local RaspberryPi and after that I want to analyze logfiles from a few remote server of the same provider. This provider stores all the logfiles in /logs

So far so good...

But... all these logfiles are zipped like: access_log_2020-03-03.gz
Is it somehow possible to read these zipped logfiles with goaccess?

Thanks in advance... :)

allinurl:
Great question. I'd use `sshfs` and mount that directory on your RaspberryPi.

JoshDebris:
Thank you for that tip.

I'm totaly overwhelmed wth the whole Linux stuff.... but I will read the documentaion about sshfs.
Hopefully I can get goaccess running at all? ;-)

allinurl:
You should be able to run it as (assuming debian on the RaspberryPi):

```
  # apt-get install sshfs
  # mkdir /home/your_user/logs
  # sshfs remote_user@remote_server.com:/var/log/apache2/ /home/your_user/logs
  # zcat -f /home/your_user/logs/access.log* | goaccess - --log-format=COMBINED
```

0bi-w6n-K3nobi:
I use `socat` for transfer and analysis in real-time.
However, you will need to have a permanent link and some mechanism/trick to deal with any failure.

allinurl:
You could try:

    # ssh -n root@server 'zcat -f /var/log/apache2/access_log*' | goaccess -

**Note**: SSH requires `-n` so GoAccess can read from stdin. Also, make sure to use SSH keys for authentication as it won't work if a passphrase is required.

PR #1869, opened 2020-08-05
README.md: Use zcat instead of pipes

alexislefebvre:
This way is simpler.

I took it in https://github.com/allinurl/goaccess/issues/1693#issuecomment-594819061


allinurl:
Merged. Thank you!

PR #1880, opened 2020-08-12
Remove license statement in Dockerfile

err931:
Removed license statement in the `Dockerfile` to conform to the primary license.


allinurl:
Merged. Thanks!

PR #1909, opened 2020-09-04
Generate random filenames for fifo files

stephenwade:
This allows multiple instances of GoAccess to run at the same time without specifying custom `fifo-in` and `fifo-out` names.

It also reads the TMPDIR environment variable if present, in case the user doesn't want the temporary files in `/tmp`.


allinurl:
Merged. Thanks for submitting this PR!

PR #1962, opened 2020-11-13
Seed random number generator

stephenwade:
#1909 used `mktemp` to generate random filenames for FIFO files. This allowed multiple instances of GoAccess to run at the same time without specifying custom `fifo-in` and `fifo-out` names.

This was changed in af76bd9bf9eb25d431b8ab279946665150b0ed0a to use custom code with `rand` to generate filenames. Unfortunately, the random number generator is never seeded, so every execution of the program uses the same filenames. This means the original problem in #1909 is now back.

This PR fixes the problem again with a different method. It seeds the random number generator with the PID of the current process. Since this is different for every process (even when several processes are started per second), this is a reliable way to generate unique FIFO names for each instance of GoAccess.

**Before:**

```
root@phobos:~# systemctl list-units 'goaccess*' --state=running --no-legend | wc -l
10
root@phobos:~# ls /tmp/goaccess* | wc -l
2
```

**After:**

```
root@phobos:~# systemctl list-units 'goaccess*' --state=running --no-legend | wc -l
10
root@phobos:~# ls /tmp/goaccess* | wc -l
20
```


stephenwade:
I also verified that this generates no collisions for any possible PID. I tested by generating a pair of FIFO names for each seed between 2 and 32768, and verified that all of the names were unique.

<details>
<summary>test_genstr.c</summary>

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include "xmalloc.h"

#define RAND_FN 7 + 1
#define PID_MAX 32768

/* A random string generator. */
void
genstr (char *dest, size_t len) {
  char set[] = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";

  for (; len > 0; --len)
    *dest++ = set[rand () % (sizeof (set) - 1)];
  *dest = '\0';
}

static char *
generate_fifo_name (void) {
  char fname[RAND_FN];
  const char *tmp = NULL;
  char *path = NULL;

  if ((tmp = getenv ("TMPDIR")) == NULL)
    tmp = "/tmp";

  memset (fname, 0, sizeof (fname));
  genstr (fname, RAND_FN - 1);

  path = malloc (snprintf (NULL, 0, "%s/goaccess_fifo_%s", tmp, fname) + 1);
  sprintf (path, "%s/goaccess_fifo_%s", tmp, fname);

  return path;
}

int
main (void)
{
  int i;

  // test all possible PIDs
  for (i = 2; i < PID_MAX; i++)
  {
    srand (i);
    printf ("seed: %d\n", i);

    // for each PID, generate a pair of FIFO names
    printf ("%s\n", generate_fifo_name ());
    printf ("%s\n", generate_fifo_name ());
  }
}
```

</details>

<details>
<summary>How I tested</summary>

```bash
$ gcc -Wall -Werror test_genstr.c -o test_genstr
$ ./test_genstr | sort | uniq --repeated
```

</details>

allinurl:
Perfect. Thanks for pointing that out! Merged.

PR #1985, opened 2020-12-26
WIP: add container based isolated build environment

thomas-mc-work:
Hello @allinurl,

I have created a way to compile the project in an isolated build environment based on Docker resp. Podman. It consists of a container image containing the environment itself, and a shell script with the actual build instructions. I've extracted the instructions and parameters from the CI config as good as possible. Could you please have a look at this and tell me 1. what you think about the whole thing and 2. whether the parameters are correct or can be improved?

Thanks in advance!


thomas-mc-work:
Oh, I forgot to mention the invocation instructions:

    # docker build -t goaccess/build.debian:1.4.3-10 -f Dockerfile.debian.10 .
    # docker run -i --rm -v $PWD:/goaccess goaccess/build.debian:1.4.3-10 > ../goaccess.bin

They are still hidden in the Dockerfile. If the idea will be accepted, then I'd remove them there and add some notes in the readme.

allinurl:
@thomas-mc-work Sorry for the delay on this and thanks a lot for submitting it. Parameters look good. Though, I'm sure I must be missing something, but can I ask, what are the benefits of having them this way? Also, please feel free to add any docs to the README.md so users can make sure how/why to use them. Thanks again!

thomas-mc-work:
What exactly do you mean? The parameters of the docker commands?

allinurl:
@thomas-mc-work sorry about that, mostly curious about having those be isolated to debian/ubuntu vs using the default Dockerfile from alpine? e.g., build size, etc. Thanks!

thomas-mc-work:
Unfortunately docker isn't available on some of the machines, that I use to run services. Or do you mean I can use the produces binary from the alpine image?

allinurl:
@thomas-mc-work You should be able to run the current dockerfile in any of your environments. Please take a look at the instructions [here](https://github.com/allinurl/goaccess#docker). Let me know if that's what you are looking to.

thomas-mc-work:
At first sorry about some confusions: I probably missed some technical background. I figured out that it's totally fine to use the compiled goaccess binary from your existing alpine Dockerfile. I thought it would be good to have the compilation process take place in the target distribution (which is Debian for me and probably many other users). So for now my approach would be useless.

But I've recognized that my binary has a weight of around 700 KB whereas the one from the alpine Dockerfile has 11 MB. I guess, this is due to the static nature of the produce triggered by the `-static` build parameter?

allinurl:
No problem at all :) Like you said, it may be worth adding them, I think nodejs offers images on top of Debian and Alpine. However, if we choose going that route, I'd like to make it work the way the [**current image**](https://github.com/allinurl/goaccess#docker) works. On the other hand, usually Alpine images come in at a light-weight, so not sure to yet.

Yeah, I agree, `-static` is probably increasing the size of the build. Do you get a similar result if you build it as static under ubuntu/debian? Thanks

thomas-mc-work:
> Do you get a similar result if you build it as static under ubuntu/debian?

I didn't manage to get it compile. It complains about missing SSL library, although `libssl-dev` has been installed. But I guess it would be bout the same size, as this is the nature of static linked binaries. They come with all libraries included.

I've also tested the dynamically linked binary produced by the existing Dockerfile on Debian. It wasn't possible to start it because it was linked against musl libc, which is absent in Debian distributions.

    $ file goaccess
    goa: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, with debug_info, not stripped

So finally having a separate build chain for Debian included in the repository would be a great benefit. Maybe the Ubuntu option is too much, as they both are like siblings.

thomas-mc-work:
If you agree to the proposal, then I'd remove the Dockerfile for Ubuntu and add the relevant documentation to the readme.

thomas-mc-work:
ping @allinurl

allinurl:
@thomas-mc-work thanks for the ping. Can you just recap what changes you're planning to commit? Thanks again!

thomas-mc-work:
Maybe now after my latest changes the goal is clearer.

> what are the benefits of having them this way?

All development dependencies are isolated in the container image. So you can quickly compile the latest version of goaccess without caring of any dependencies except for docker/podman.

> please feel free to add any docs to the README.md

That's done now.

> mostly curious about having those be isolated to debian/ubuntu vs using the default Dockerfile from alpine?

If you have a container runtime on the target system then you'll be fine just using the provided Alpine image. On my target system, I can only use binaries located within the users home folder, so no access to `apt`. That's why I'm looking for an official way to build the binary for Debian locally (has to be a separate machine with Docker of course).

> e.g., build size

There shouldn't be any reason for a difference.

---

Finally this PR is only about the compilation process. So when compared to the existing Dockerfile, then only the first half of it can be counted.

allinurl:
Merged. Thanks a lot for submitting this PR!

PR #1989, opened 2021-01-01
configure: Work on NetBSD

rsmarples:
We need to link to libintl to get dgettext.
We need to link to curses to get refresh.
test == is not portable, test = is.

Signed-off-by: Roy Marples <roy@marples.name>


allinurl:
Thanks very much. This has been merged!

PR #1990, opened 2021-01-02
POSIX has made usleep(2) obsolete in favour of nanosleep(2)

rsmarples:
usleep(2) also has no effect if equal to greater than 1000000
which happens to match HTML_REFERS, but no error was checked.
This caused 100% CPU utiliation constantly following the tail
of the web server access log with default value.

nanosleep(2) has more relaxed limits.
This commit does a like for like replacement and also adds
error checking in case of a similar error in the future.

Signed-off-by: Roy Marples <roy@marples.name>

### DIFF
--- a/src/goaccess.c
+++ b/src/goaccess.c
@@ -845,8 +845,11 @@ perform_tail_follow (GLog * glog) {
 out:
 
   if (!conf.output_stdout) {
+    struct timespec ts = { .tv_sec = 0, .tv_nsec = 200000000 }; /* 0.2 seconds */
+
     tail_term ();
-    usleep (200000);    /* 0.2 seconds */
+    if (nanosleep(&ts, NULL) == -1 && errno != EINTR)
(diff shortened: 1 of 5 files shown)


allinurl:
Merged. Thanks very much for the heads up on this!

allinurl:
Question, were you running goaccess from systemd?

rsmarples:
Well, as my prior pull request fixed the build for NetBSD that can safely be a no as systemd is Linux only.

However, I did see the issue you were referring to and all I can say is fix this is highly dependant on libc.
Internally all modern libc will use nanosleep when usleep is called.
However, for NetBSD at least there is an extra check to ensure that EINVAL is returned for >= 1000000 as POSIX dictates whereas glibc and musl at least do not - hence non BSD users don't have much of an issue with this.

allinurl:
Got it. Thanks for the info!

PR #1993, opened 2021-01-06
Fix build when clang is the compiler

rsmarples:
While here, respect the host default compiler of cc.
If that fails, then try gcc and finally clang.

Now I can do
./configure
make

and there are zero warnings when cc is clang on NetBSD.


allinurl:
Merged. Thank you for fixing these!

PR #1994, opened 2021-01-06
Detect SHA1Init in libc and use that instead of our own sources

rsmarples:
This allows goaccess to be built with address sanitisation on NetBSD, otherwise it would not know which SHA1 functions to use.


allinurl:
Merged. Thank you again!

PR #2007, opened 2021-01-26
Monitor log files using kqueue(2)

rsmarples:
This patch replaces the behaviour of opening, checking,
processing and closing each logfile every few moments with a
kqueue(2) based file monitor.

The new behaviour is similar to poll(2) in that the OS will inform us
when the file has changed. To do this, we need to hold a descriptor
open at all times rather than close it between runs.
If the logfile is turned over, we then monitor the directory for
any modifications in the same way and on a change we try to re-open
the logfile once more.

This slightly reduces the CPU time goaccess spends on NetBSD at least,
but is a massive saving on the CPU time the kernel spends constantly
stating, opening and closing the logfile.


rsmarples:
This has only been tested on NetBSD. Linux could get a similar implementation using epoll.

allinurl:
Thanks for submitting this. I always thought it would be great to have this change upstream, either through `kqueue` as you did or `epoll`. 
Would you be able to take a look at what might be the issue, it seems like it doesn't like when piping data in and outputting to a real-time html report. e.g.,

    ssh -n root@server 'tail -f /var/log/apache2/access.log' | ./goaccess - --log-format=COMBINED -o report.html --real-time-html

or with two logs, local + pipe

    ssh -n root@server 'tail -f /var/log/apache2/access.log' | ./goaccess /var/log/apache2/local-access.log - --log-format=COMBINED -o report.html --real-time-html

Thank you!

rsmarples:
Should work now, but not tested.

allinurl:
Would you be able to run the following command on your end, I get a segfault which might be related to `glog->fp` being null due to `-`. 

    # goaccess - access.log --log-format=COMBINED

Thanks!

rsmarples:
> Would you be able to run the following command on your end, I get a segfault which might be related to `glog->fp` being null due to `-`.
> 
> ```
> # goaccess - access.log --log-format=COMBINED
> ```

Works locally here.

allinurl:
I wonder if it has something to do with me testing it on freebsd.

rsmarples:
Good to know that goaccess works with NetBSD curses (hint, it's not ncurses).
Your problem should be fixed now.

allinurl:
Merged. Thank you again for addressing those issues and submitting this!

rsmarples:
Would you reconsider if I added Linux inotify(7) support to match kqueue?
The implementation would be quite similar and would share the majority of the code path.

rsmarples:
Or at least hide it away under a configure switch so I can use it :)

allinurl:
Before rushing out on my response, let me think about these two options and I'll get back to you.

PR #2009, opened 2021-01-26
Support real-time piping as non-root user

### LINKED ISSUE #868: Real-time piping as non-root never finishes parsing
I'm on 7.3 running GoAccess 1.2.

Piping log files like: `tail -f access.log | goaccess -` works ok as root, however when executed as non-root, goaccess is stuck at "Parsing...". It parses all the lines but does not seem to know that the "end" was reached.

I've tried looking if anyone has the same problem but I can't find any.

Schnouki:
If running as a user that can't open a terminal, it's still important to set the
pipe as non-blocking. Otherwise parsing never finishes, making goaccess
unusable.

Fixes #868.

### DIFF
--- a/src/goaccess.c
+++ b/src/goaccess.c
@@ -1244,21 +1244,24 @@ static FILE *
 set_pipe_stdin (void) {
   char *term = NULL;
   FILE *pipe = stdin;
-  int fd1, fd2;
+  int term_fd = -1;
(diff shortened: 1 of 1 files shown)


allinurl:
Merged - awesome, thanks a lot for submitting this PR!

PR #2034, opened 2021-02-20
add some browsers

IzzySoft:
This massively reduces the "unknown" entries in the browser section of my reports.

* Waterfox & PaleMoon are Firefox forks, so I assigned them to "Firefox"
* Pleroma is used by the decentralized, mastodon compatible Pleroma (when fetching "link previews" for the timeline I guess), Brainstorm seems to be a bot fetching content for Reddit, Mastodon & Matrix – hence I assigned both to "Feeds"
* F-Droid is used by the official [F-Droid](https://f-droid.org/) Android client, okhttp by several Android apps – hence assigned to "Others"

"F-Droid" will probably only show up on servers providing F-Droid app repositories. I'm running such a repo, hence I'm interested in having it defined – but if that's not wanted "globally" I fully understand and would "patch" it on my end before compile, just let me now if you want it removed.

### DIFF
--- a/src/browsers.c
+++ b/src/browsers.c
@@ -122,6 +122,8 @@ static const char *browsers[][2] = {
   {"libwww-perl", "Others"},
   {"python-requests", "Others"},
   {"PackageKit", "Others"},
+  {"F-Droid", "Others"},
+  {"okhttp", "Others"},
 
   /* Feed-reader-as-a-service */
   {"AppleNewsBot", "Feeds"},
@@ -138,6 +140,8 @@ static const char *browsers[][2] = {
   {"theoldreader.com", "Feeds"},
   {"WordPress.com Reader", "Feeds"},
   {"YandexBlogs", "Feeds"},
+  {"Brainstorm", "Feeds"},
+  {"Pleroma", "Feeds"},
 
   /* Google crawlers (some based on Chrome,
    * therefore up on the list) */
@@ -152,6 +156,8 @@ static const char *browsers[][2] = {
   /* Rebranded Firefox but is really unmodified
    * Firefox (Debian trademark policy) */
   {"Iceweasel", "Firefox"},
+  {"Waterfox", "Firefox"},
+  {"PaleMoon", "Firefox"},
   {"Focus", "Firefox"},
   /* Klar is the name of Firefox Focus in the German market. */
   {"Klar", "Firefox"},


allinurl:
Awesome! Thank you for submitting this PR. Merged.

IzzySoft:
That was fast! OK, then I need to send my update with a separate MR :smile:

Just a question: locally, I also updated `opesys.c` with strings where the UA doesn't mention the OS but clearly is available for a specific US only. As I saw no such strings in that file I'm unsure if that's wanted, especially with the "slow" note on top?

allinurl:
Good question. There are a lot of browsers and bots that are certainly not in that list which I could have included. It is that way explicitly so goaccess doesn't slow down upon parsing the log. My intent was to have a basic but well-curated list of browsers and OSs and let users use the `browser.list` file to locally add additional ones depending on the traffic their server gets. Especially since not everyone will receive the same bots/traffic.

Maybe I can add a way to dump a list of all the UAs that were identified as `Unknown` so that users can add those UA identifiers to the `browsers.list`. Thoughts?

IzzySoft:
Yes, that intent was understood (actually I half expected you'd want me to drop some of the UAs I've added with this or my follow-up PR (which is still open btw, see #2035), so I gave some background on each to ease that decision).

Dumping a list of `Unknown` would be really helpful, I had already looked for that in vain :smile: Same for OS btw, if that's possible at all – not all UA strings even mention an OS, though some are easily deducted when a given UA is only available for a specific OS (like `"F-Droid x.yz"` with `x.yz` being a version number always means Android). Then again, you'll see that UA only when you're serving an F-Droid repo – which is why I didn't submit that PR. For another (`okhttp`) I'm not 100% sure if it always indicates Android or could well be some other Java application. Then again, seeing all UAs leading to `Unknown` OS might help compiling a list.

allinurl:
I think the ones you submitted are fairly common to make it to the list. Though, at some point I'd need to trim the list a bit as it has been growing lately and try to keep really the most common ones in the `browsers.c` source file.

I agree, I think that's why there is a substantial amount of unknowns in the OS panel, for instance I think many bots don't list an OS as you mentioned, e.g., `bingbot` so maybe we should just list them as Crawlers in the OS list? On the other hand, I think if we can determine the OS, e.g., `F-Droid`, we should list them as such.

I'll work on adding that option for the upcoming version.

IzzySoft:
> so maybe we should just list them as Crawlers in the OS list?

Sounds reasonable on first look – but then, this must be consequent. And there it becomes a question: what do you do if you know that Crawler always runs on a specific OS, or the UA string explicitly says which it is? Then you have some Crawlers as Crawlers, and others with their specific OS – which might be a bit confusing. So maybe "Crawlers with unknown OS" or something along these lines? ("Unknown (Crawlers)" would be shorter). Then maybe also rename "Unknown" to "Unknown/others")?

So yes, if we can determine the OS, this would certainly be preferable. Question is just how much that affects speed. I can do some measurements here, logs of my most frequented sites are 30 (Android topics), 50 (eBooks), 250M (repositories: APT, YUM, F-Droid) and with the latter, a full scan noticeably takes some time. So I could e.g. build 2 variants of GoAccess and measure the runtime differences. Only tricky part: after the first scan, the file might be cached by the OS, so subsequent scans are noticeably faster :wink:

allinurl:
I pushed a [commit](https://github.com/allinurl/goaccess/commit/ef593116b336253d3c5cdfc1a736ecdda3f75104) that should let you dump a list of unknown browsers and OSs. The option is `--unknowns-log=<filename>`. Feel free to build from [**development**](https://github.com/allinurl/goaccess#build-from-github-development) to test it out. It will be pushed out in the upcoming version. Note that if you have a lot of unknowns and your log is huge, it may take longer to parse & dump the file. Then you can sort as:

    # cat unknowns | sort | uniq -c | sort -nr > unknowns-sorted

To get a list that looks like:

```
  96585 [OS]   -
  96585 [BR]   -
  36314 [OS]   Mozilla/5.0 (compatible; MJ12bot/v1.4.5; http://www.majestic12.co.uk/bot.php? )
  33546 [OS]   Mozilla/5.0 (compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)
  13013 [OS]   Mozilla/5.0 (compatible; Yahoo! Slurp; http://help.yahoo.com/help/us/ysearch/slurp)
   8955 [OS]   Mozilla/5.0 (compatible; Baiduspider/2.0;  http://www.baidu.com/search/spider.html)
   8151 [OS]   Mozilla/5.0 (compatible; spbot/5.0.1;  http://OpenLinkProfiler.org/bot )
   7776 [OS]   Mozilla/5.0 (compatible; MegaIndex.ru/2.0;  http://megaindex.com/crawler)
   6516 [OS]   msnbot-media/1.1 ( http://search.msn.com/msnbot.htm)
   6516 [BR]   msnbot-media/1.1 ( http://search.msn.com/msnbot.htm)
   4456 [OS]   Mozilla/5.0 (compatible; DotBot/1.1; http://www.opensiteexplorer.org/dotbot, help@moz.com)
   4160 [OS]   Mozilla/5.0 (compatible; AhrefsBot/5.1;  http://ahrefs.com/robot/)
   3888 [OS]   Mozilla/4.0 (compatible; Synapse)
   3786 [OS]   Mozilla/5.0 (compatible; YandexBot/3.0;  http://yandex.com/bots)
   3342 [OS]   facebookexternalhit/1.1 ( http://www.facebook.com/externalhit_uatext.php)
   3282 [OS]   Mozilla/5.0 (compatible; BLEXBot/1.0;  http://webmeup-crawler.com/)
   2679 [OS]   Mozilla/5.0 (compatible; XML Sitemaps Generator; https://www.xml-sitemaps.com) Gecko XML-Sitemaps/1.0
   2451 [OS]   TwitterFeed 3
   2365 [OS]   Aboundex/0.3 (http://www.aboundex.com/crawler/)
   ...
```

I agree, in that case, like you said, changing the category name would make more sense. I'd be curious to get some numbers if you can share those :)

IzzySoft:
Yes, most of those apps I see here in my logs as well. Question probably is the balance again: what is the threshold here? Just add any bot we encounter – or keep rather minor ones out for performance reasons? Taking your above list, frequent visitors here match…

* Spiders: MJ12bot, Yahoo! Slurp (maybe shorten to "Yahoo" to also cover their images bot?), Baiduspider, DotBot, AhrefsBot, YandexBot
* Others: Synapse (maybe "Feed"? It seems to be the Matrix server, so it would be a similar category as Mastodon/Pleroma)

The other ones do not sound familiar to me, but I'd need to first create the list to be sure. Before I do so, two questions:

* building from Git mentions one additional command: `autoreconf -fiv`. There's no `autoreconf` in my `$PATH`. Which Debian package would ship that? My guess is `autoconf`?
* with my last build (from release), after the `make` I simply replaced the binary. Didn't dare running `make install` as I first wanted to see if it works at all. Wouldn't `make install` also overwrite my config etc? Sure, I could simply backup and restore it. But what would be the recommended way to update an existing installation?
* oh, another one: release is currently 1.4.5 – while the latest tag is 1.4.4; a bit irritating :wink:

PR #2038, opened 2021-02-23
Allow goaccess to be cross compiled.

ericonr:
When cross compiling, bin2c needs to be built for the build system, so
building it with the CC compiler won't work. We use an autoconf macro in
order to obtain a good value for CC_FOR_BUILD, and define a rule using
it directly instead.

Reference
https://www.gnu.org/software/autoconf-archive/ax_cc_for_build.html

---

If you want to avoid the dependency on `autoconf-archive` for the new macro, the logic is reasonably easy to replicate.

I noticed that `bin2c.c` has some code for using bzip2, but I don't think that can be enabled in any way, at least for now?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,8 +12,8 @@ dist_noinst_DATA =                 \
   resources/js/d3.v3.min.js        \
   resources/js/hogan.min.js
 
-noinst_PROGRAMS = bin2c
-bin2c_SOURCES = src/bin2c.c
+bin2c$(EXEEXT): src/bin2c.c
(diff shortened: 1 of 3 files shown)


ericonr:
CI seems to be failing because `CC_FOR_BUILD` isn't defined, which I would assume happened because it lacks `autoconf-archive`.

allinurl:
Thanks for submitting this and sorry for the delay. Are you able to avoid the `autoconf-archive` dependency? I'd certainly like to it keep without that dependency.

Yeah I did not include the compression part as an option. Do you need it in your case?

Thanks again!

Issue #2071, opened 2021-04-03
GNU+Linux incorrectly referred to as "Linux"


throwaway1037:
GoAccess incorrectly refers to the GNU+Linux operating system as "Linux" in its user interfaces and source code.

GNU is an OS started in 1983 and Linux is a kernel started in 1991. The combination of these two components today forms the GNU+Linux operating system.

See also:
https://www.gnu.org/gnu/linux-and-gnu.html
https://www.gnu.org/gnu/gnu-linux-faq.html
https://www.gnu.org/gnu/why-gnu-linux.html

allinurl:
Please feel free to submit a PR with the relevant changes. Thanks

allinurl:
Merged. Thanks!

throwaway1037:
:)

PR #2072, opened 2021-04-03
Fix naming

throwaway1037:
As per issue #2071, the GNU+Linux operating system is now referred to correctly. The name "Linux" was retained only when used correctly; when it was referring to the Linux kernel and not the GNU+Linux operating system.

I also renamed a couple of files as markdown files for compatability, but then changed them back because the autotools are dumb.


allinurl:
Merged. Thank you!

throwaway1037:
:)

PR #2074, opened 2021-04-04
Change to neutral term

throwaway1037:
"Open source" is a term used to show support for open source, whereas "free" or "libre" are terms used to show support for the free software movement.

I assume goaccess wants to remain neutral, so I switched all mentions of "open source" to the neutral term "FLOSS" (free (libre) and open source software).

Please see also:
https://www.gnu.org/philosophy/words-to-avoid.html#FLOSS
https://www.gnu.org/philosophy/floss-and-foss.html

### DIFF
--- a/README
+++ b/README
@@ -1,6 +1,6 @@
 What is it?
 -------------
-GoAccess is an open source real-time web log analyzer and interactive viewer
+GoAccess is a FLOSS real-time web log analyzer and interactive viewer
 that runs in a terminal in *nix systems or through your browser.
 
 It provides fast and valuable HTTP statistics for system administrators that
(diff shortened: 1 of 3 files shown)


throwaway1037:
Personally, I would recommend ditching the idea of open source and switching to free (libre) software, although I presume the goaccess community would prefer the neutral term "FLOSS".

dertuxmalwieder:
Don't expect the audience of GoAccess to understand the term "FLOSS" without an explanation. "Free (libre)" would make more sense than "FLOSS".

allinurl:
Thanks for submitting this. To be honest, as @dertuxmalwieder mentioned, I think most folks won't understand FLOSS. Also, isn't the free/libre part somehow dictated by the license?

PR #2076, opened 2021-04-04
Don't call unknown Linux systems "GNU+Linux"

### LINKED ISSUE #2075: Non-GNU Linux systems are mislabeled as GNU+Linux recently
In 2e15d8ef21edc5952ba6ba3ad321b5976a165861, several pull requests from @throwaway1037 were merged which included renaming "Linux" to "GNU+Linux". Now that would mean that all Linux systems which are detected are GNU systems instead of (e.g.) Busybox-based Linuces, like Tiny Core Linux and Alpine Linux.

As there is no safe way to detect these libc-related differences, the commit - in opposite to what it *should* have done - has made the OS detection **less** correct under some circumstances, because all Linux systems use the Linux kernel, but not all Linux systems use the GNU userland.

I suggest to undo this part of the commit.

dertuxmalwieder:
Should fix #2075

### DIFF
--- a/src/opesys.c
+++ b/src/opesys.c
@@ -99,8 +99,8 @@ static const char *os[][2] = {
   {"OpenBSD", "BSD"},
   {"DragonFly", "BSD"},
 
-  {"Linux", "GNU+Linux"},
-  {"linux", "GNU+Linux"},
+  {"Linux", "Linux"},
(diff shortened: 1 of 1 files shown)


allinurl:
Makes sense. Thanks for pointing that out!

PR #2077, opened 2021-04-06
Correct system classification

throwaway1037:
GNU+Linux is a sane default, as most systems using the Linux kernel are GNU+Linux.

Add a couple of exceptions to the rule of thumb.

Add a few more GNU+Linux distros.

### DIFF
--- a/src/opesys.c
+++ b/src/opesys.c
@@ -85,6 +85,7 @@ static const char *os[][2] = {
   {"Ubuntu", "GNU+Linux"},
   {"Fedora", "GNU+Linux"},
   {"Mint", "GNU+Linux"},
+  {"OpenSUSE", "GNU+Linux"},
   {"SUSE", "GNU+Linux"},
   {"Mandriva", "GNU+Linux"},
   {"Red Hat", "GNU+Linux"},
@@ -92,15 +93,27 @@ static const char *os[][2] = {
   {"CentOS", "GNU+Linux"},
   {"PCLinuxOS", "GNU+Linux"},
   {"Arch", "GNU+Linux"},
-  {"Parabola", "GNU+Linux"},
+
+  /* Entirely libre systems */
+  {"Parabola", "GNU+Linux-libre"},
+  {"GNU Guix", "GNU+Linux-libre"},
+  {"PureOS", "GNU+Linux (deblobbed)"},
+  {"Trisquel", "GNU+Linux (deblobbed)"},
 
   {"FreeBSD", "BSD"},
   {"NetBSD", "BSD"},
   {"OpenBSD", "BSD"},
   {"DragonFly", "BSD"},
 
-  {"Linux", "Linux"},
-  {"linux", "Linux"},
+  /* Most systems using the Linux kernel are
+   * GNU+Linux, with the exception of the Android
+   * family of systems, listed below, and a few
+   * other distributions, listed here. */
+  {"TinyCore", "BusyBox+musl+Linux"},
+  {"Alpine", "BusyBox+musl+Linux"},
+
+  {"Linux", "GNU+Linux"},
+  {"linux", "GNU+Linux"},
 
   {"PlayStation", "BSD"},
 


emilengler:
Your approach is kludge. It's not the task of the repository to maintain a list of GNU/Linux and non GNU/Linux distros. Also as far as I know this program doesn't depend or uses any GNUisms, just the Linux kernel (through syscalls). Plus your approach has failed. I know at least one other distro that is not GNU/Linux which you haven't listed meaning it doesn't fix anything.

dertuxmalwieder:
I agree with @emilengler here, even although I usually disagree with him. You will inadvertently mislabel non-GNU systems as GNU with this approach.

throwaway1037:
With your approach, you will consistently and more frequently mislabel GNU.

It's also irrelevant whether goaccess depends on GNU or not (it does by the way); that's still the name of the OS.

emilengler:
> With your approach, you will consistently and more frequently mislabel GNU.

It **HURD**s me and is also wrong that GNU/Linux is considered as a full GNU system.

> It's also irrelevant whether goaccess depends on GNU or not (it does by the way); that's still the name of the OS.

Nonetheless Linux remains the lowest common denominator. Every GNU/Linux system is a Linux system but not every Linux system is a GNU/Linux system.

dertuxmalwieder:
Naming a GNU+Linux system “Linux” is correct as it uses the Linux kernel. Maybe a better approach with less false positives would be to rename “Operating Systems” “kernels” or something? (The BSDs are likewise put into one category.)

throwaway1037:
Why do you never learn?

***THERE IS NO SUCH THING AS A "LINUX SYSTEM"!***

What do you have against GNU?

For the 100th time, 99.9% of systems using the Linux kernel are GNU+Linux, with the exception of the Android family of systems, and the few tiny BusyBox+musl+Linux edge-cases like TinyCore and Apline.

dertuxmalwieder:
> ***THERE IS NO SUCH THING AS A "LINUX SYSTEM"!***

***NOT ALL LINUX KERNEL BASED SYSTEMS ARE GNU!*** Why do you never learn?

> 99.9% of systems using the Linux kernel are GNU+Linux

Which adds 0.1% false positives. No, thank you.

throwaway1037:
We're going round in circles. I'm the only one who has committed a kludgy albeit functional implementation with a 0.01% false positive rate.

Your proposed system of labelling everything "Linux" is wrong 99.9% of the time, unless the 0.01% edge-case is hit. Your system comes broken.

throwaway1037:
Just because something uses the Linux kernel does not make the whole system "Linux". This meme has been going on for too long, now. Give GNU the credit they deserve.

dertuxmalwieder:
> I'm the only one who has committed a kludgy albeit functional implementation

... of **knowingly** mislabeling operating systems. Golly gee, what a fine thing you did!

> with a 0.01% false positive rate.

0.1.

> Your proposed system of labelling everything "Linux" is wrong 99.9% of the time

I think most people here will agree that a system which uses the Linux kernel can safely be named as "using the Linux kernel". What you **mean** is that the term "operating system" is wrong, but what you **do** is breaking the accuracy of our website statistics for pseudo-religious reasons.

> Just because something uses the Linux kernel does not make the whole system "Linux". 

Which is why Android is categorized separately.

> Give GNU the credit they deserve.

RMS will agree that listing non-GNU systems as "GNU" is the wrong thing to do, yet you insist to do that.
It is **just wrong** to assume that Linux kernel based systems (or whatever is your preferred term for that) are "GNU+Linux systems" unless they declare so.

dertuxmalwieder:
I mean, you could also say "GNU+Windows" to Windows 10 just because it comes with an (optional) GNU environment. Waiting for your pull request...

throwaway1037:
I steak great offence at you labelling me "pseudo-religious". I do not believe in religion. Why would you say that?

The truth is that you just don't want to admit that the system is actually GNU so you will go to great lengths to avoid saying or mentioning it to save face.

My implementation is incorrect 0.01% of the time; yours is wrong 99.9% of the time. I fail to see how your implementation can possibly be superior.

I absolutely **do not** want to mislabel non-GNU systems as GNU and I assert that I never knowingly did do, hence I added the edge-cases explicitly. I did exactly what you wanted yet that is somehow not good enough.

throwaway1037:
I know you are winding me up with the Windows+GNU thing. This shows that you fundamentally misunderstand *why* the system is called GNU+Linux in the first place.

Please read the articles on the GNU website about Linux and the GNU system, and the GNU+Linux FAQ.

Nevertheless, the Subsystem for GNU+Linux on Windows should be named as such.

dertuxmalwieder:
> The truth is that you just don't want to admit that the system is actually GNU

Not all systems which use the Linux kernel are actually GNU, so please stop trying to break the reliability of the website statistics by mislabeling them.

throwaway1037:
But don't you see that you are doing the very same thing by labelling GNU+Linux as "Linux"?

dertuxmalwieder:
> But don't you see that you are doing the very same thing by labelling GNU+Linux as "Linux"?

100% of all systems which use the Linux kernel use the Linux kernel.

throwaway1037:
True, but you're labelling all systems using the Linux kernel as "Linux" when 99.9% of them, with the exception of the Android family of systems, are actually GNU+Linux.

dertuxmalwieder:
I am still waiting for your code that **safely detects** GNU and non-GNU systems without keeping a complete list of Linux distributions.

throwaway1037:
If we can only name three non-GNU systems using the Linux kernel, then I think that proves my point. There are thousands of GNU+Linux distros.

throwaway1037:
Explicitly listing them is the only way (at least for now).

throwaway1037:
Theres no such thing as a "Linux distribution". Linux is a kernel, not an OS. This confusion is why I added GNU+Linux in the first place.

dertuxmalwieder:
> Explicitly listing them is the only way (at least for now).

Please show me a complete list of GNU and non-GNU Linux distributions which does not contain any false positives (because false positives make website statistics useless). Include LFS in your considerations.

> Linux is a kernel, not an OS. 

Rename "operating systems" into "operating systems and kernels". Problem solved. Stop mislabeling non-GNU operating systems.

emilengler:
> Explicitly listing them is the only way (at least for now).

The only solution I know is the following:

1. Compile a C program which checks the macro `__GLIBC`
2. Execute the C program
3. Depending on it's output, fill the .go file
4. Compile the Go program

but is that worth it...

throwaway1037:
No, don't conflate OSs and kernels. That is partly why this confusion exists at the moment.

throwaway1037:
How would goaccess exactly run uname on a remote machine via reading a log file?

dertuxmalwieder:
> No, don't conflate OSs and kernels.

The kernel is the only part of a GNU/Linux or non-GNU/Linux system which GoAccess can detect without a single false positive.

allinurl:
I do appreciate the discussion about the proper labeling of OSs. However, after thinking about this, I'm not sure the majority of goaccess' users are interested or would like to see a bunch of different categorizations that for most folks are mostly known as Linux. 

Nevertheless, I'm aware of the existing naming controversy and the improper use of simply using Linux for all distros out there, but again, I think it would be a big turnoff for a lot of people that use goaccess if they start to see all these different categories.

I personally think it would make more sense to fix it directly at the distro level and the user agent it sends. For instance,

    Mozilla/5.0 (X11; Debian; Linux x86_64; rv:42.0) Gecko/20100101 Firefox/42.0 

perhaps should be

    Mozilla/5.0 (X11; Debian GNU/Linux; Linux x86_64; rv:42.0) Gecko/20100101 Firefox/42.0

Also, I don't even know if a lot of the distros listed on this commit send the distro name on it. If they do, please feel free to post those user agent strings.

For now I'm closing this, but once again I appreciate the note and the PR.

PR #2078, opened 2021-04-06
Switch to libre OS

throwaway1037:


dertuxmalwieder:
Why?

allinurl:
I'll close this one as there's no much background as why the change. Alpine works fine in this case.

PR #2079, opened 2021-04-06
Add a new category for Unix

dertuxmalwieder:
Solaris/SunOS/illumos is an actual UNIX, not just "Unix-like"; also, added AIX (which is still alive).


allinurl:
Merged. Thank you!

PR #2102, opened 2021-05-06
Rename referer to referrer in goaccess.conf

dertuxmalwieder:
Now that GoAccess prefers "referrer" over "referer", the example configuration file should probably cover that.


allinurl:
Merged. Thank you!

dertuxmalwieder:
Looking forward to the next update I can break!

Issue #2103, opened 2021-05-06
The d3 library is old and could need an update


dertuxmalwieder:
As far as I can see, GoAccess uses version 3 of `d3.js` (the current version is [6.7.0](https://github.com/d3/d3/releases)). While this does not automatically make it a bad choice, it is probably a good idea to check whether an update (admittedly, involving a migration 3 -> 4 -> 5 -> 6, due to the API changes) could bring any advantages.

`d3.v6.min.js` weighs 264 Kb, while `d3.v3.min.js` weighs "only" 148 Kb, which is the only obvious disadvantage to me; but then again, I have not evaluated GoAccess's usage of the graphing libraries well enough to understand what else could be a problem here. I'll just leave this here as a suggestion.

allinurl:
I agree. I've been thinking to upgrade as well. Though, last time I checked, it would require some changes under [charts.js](https://github.com/allinurl/goaccess/blob/master/resources/js/charts.js). It'd be nice if we could make the charts.js code smaller if v6.7.0 helps to do that, it'd make up for the larger size of d3.

dertuxmalwieder:
I would guess that the migration guides would cover the required changes?

allinurl:
I think so, I'd have to look into it again. By the way, are you familiar with d3?

dertuxmalwieder:
Not at all, I’m afraid. There are things even I cannot understand yet.

allinurl:
As of this post, D3 was updated to v7 (latest).

PR #2110, opened 2021-05-14
D3 update

jd1048576:
This is an attempt to update d3 to address #2103. The d3 min bundle could be made smaller, as all the packages are now modular so the bundle could be changed to just use the required modules but I have left that change out for now. I have noticed some strange behavior when the page first loads the graphs don't seem to immediately show without scrolling, I am not sure what is causing this, any suggestions would be appreciated.


allinurl:
Thanks for submitting this. Any reason for closing this out?

jd1048576:
It got more complicated than I expected and unfortunately I don't have much time to work on this now, I believe the issues I was experiencing with rendering were due to the fact that how d3 handles updates to selections has changed https://github.com/d3/d3/blob/master/CHANGES.md#selections-d3-selection

jd1048576:
@allinurl I had updated my branch https://github.com/allinurl/goaccess/compare/master...jd1048576:d3-update, but it won't let me reopen the pull request, I forgot to push my attempts to address the selection update issue and an attempted at a reduced bundle. See here for the extra 2 commits, it also seems like that its not handling zero properly 
image
If you need me to do anything else just let me know.

PR #2125, opened 2021-05-29
Fix out-of-tree builds

vincentbernat:
Out-of-tree build was fixed in 51632aaedafc, but for some reason, I
have missed an occurrence.


allinurl:
Merged. Thank you!

Issue #96, opened 2014-03-27
Ability to specify search engine query


allinurl:
Allow other search engines other than Google, and let the user to specify the query parameter.

PR #2161, opened 2021-08-07
Enable local site-search and support for lowercasing searches

atiro:
A rough start at supporting local site searches (ala #96 for other search engines) , that are making use of the same query param ('q') as Google search.

Not sure if would be better as a new output instead of overriding the Google table ?

Have also added downcasing of the search term which sometimes makes results more useful for us. This might be a niche concern.


allinurl:
Thanks a lot for submitting! Question, would you be able to add the ability for the user to specify the query param? It would make this a lot more flexible.

atiro:
Good idea, working on it!

atiro:
Updated version @allinurl but not entirely sure I'm taking the best approach to save the param variations in config options (really I need to add two more as well to handle encoded). Any better option ?

PR #2182, opened 2021-09-13
Fix minor time-of-use / time-of-check "bug" in src/parser.c

kokke:
This PR will fix a tiny 'problem' with checking pointer-validity "too late", i.e. after it is already dereferenced.

See https://github.com/allinurl/goaccess/blob/master/src/parser.c#L435

Problem:  pointer 'req' is dereferenced (at line 437 by `strlen(req)`) before being checked against NULL (at line 441)
Solution: move check up before use, e.g. move if-stmt at line 441 to line 437 so that this code

```C
  435  static int
  436  verify_static_content (const char *req) {
  437    const char *nul = req + strlen (req); // <-- dereferences pointer 'req'
  438    const char *ext = NULL, *pch = NULL;
  439    int elen = 0, i;
  440  
  441    if ((req == NULL) || (*req == '\0'))  // <-- check validity of pointer 'req' (after it was dereferenced)
  442      return 0;
```

... becomes

```C
  435  static int
  436  verify_static_content (const char *req) {
  437    if ((req == NULL) || (*req == '\0')) // <-- check validity of pointer 'req' before it gets dereferenced
  438      return 0;
  439
  440    const char *nul = req + strlen (req);
  441    const char *ext = NULL, *pch = NULL;
  442    int elen = 0, i;
```

Otherwise, if `req == NULL` the program will crash/segfault at the `strlen(req);` before reaching the if-statement and returning 0.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -434,12 +434,12 @@ clean:
  * On success, the 1 is returned. */
 static int
 verify_static_content (const char *req) {
+  if ((req == NULL) || (*req == '\0'))
+    return 0;
+
   const char *nul = req + strlen (req);
   const char *ext = NULL, *pch = NULL;
   int elen = 0, i;
-
-  if ((req == NULL) || (*req == '\0'))
-    return 0;
 
   for (i = 0; i < conf.static_file_idx; ++i) {
     ext = conf.static_files[i];


allinurl:
Merged. Thanks a lot for catching this and submitting the PR!

Can I ask, did you find it as a result of triggering an error?

kokke:
Hi @allinurl and thanks for your quick feedback!

I'm sorry I left that out. This was found with a static analysis tool - no actual errors were triggered. That's also the reason I put "bugs" and "problem" in quotation-marks. 

The tool flags places where a pointer is checked against NULL, but after it has already been dereferenced. 

The dereference implies the programmer thinks the pointer is != NULL, and the NULL-check implies the opposite: both can't hold true.

The code would crash if passed a null-pointer, but maybe that situation cannot happen. I have not verified that for certain.

allinurl:
Got it. Thanks again for fixing this! I plan to release a new version with this fix on it. Stay tuned!

PR #2196, opened 2021-09-29
Docker realtime requires tail instead of cat

tofra:
Realtime requires the usage of tail instead of cat. Otherwise there are no updates


allinurl:
Merged. Thanks!

PR #2213, opened 2021-10-22
Add more useragents!

Titaniumtown:
Here's a small PR with some useragents I found in my logs, I looked up what they were and put them in the correct categories.

### DIFF
--- a/src/browsers.c
+++ b/src/browsers.c
@@ -127,6 +127,9 @@ static const char *browsers[][2] = {
   {"okhttp", "Others"},
   {"node", "Others"},
   {"PrivacyBrowser", "Others"},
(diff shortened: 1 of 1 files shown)


allinurl:
Merged. Thank you!

At some point we will need to start curating the default list as performance could degrade overtime. Essentially shoot for the ones that are the most common regardless of the type of traffic and either let users manage their custom list or somehow have goaccess automatically add them as unknowns and display those.

By the way, if you haven't use `--unknowns-log=<filename>`, it may help to identify the missing ones.

Titaniumtown:
> By the way, if you haven't use `--unknowns-log=<filename>`, it may help to identify the missing ones.

I did not know of that, thank you!

PR #2221, opened 2021-11-01
Fix mistake on "Linux"

windyOC:
I don't know why this mistake never be found.....


allinurl:
Thanks for pointing that out. *nix is actually okay, more info on this at this [post](https://en.wikipedia.org/wiki/Unix-like).

PR #2248, opened 2021-12-13
Misc

cgzones:
* Drop unicode character from status description

* Mark djb2 parameter const

  Improve const correctness by declaring the read-only parameter to djb2 const.

* Include config.h to define HAVE_LIBSSL

  It seems to work so far cause all translation units using any of the HAVE_LIBSSL dependent fields are including config.h before websocket.h.


allinurl:
Thanks a lot for submitting this! Merged.

PR #2266, opened 2022-01-20
Add additional bots to crawlers

HannesHofer:
add various additional crawlers/bots:

DataForSeoBot - https://dataforseo.com/dataforseo-bot
SeznamBot - http://napoveda.seznam.cz/en/seznambot-intro/
coccocbot - http://help.coccoc.com/searchengine
Neevabot - https://neeva.com/neevabot
Mail.RU_Bot - http://go.mail.ru/help/robots
CCBot - http://go.mail.ru/help/robots
Cincraw - http://cincrawdata.net/bot/

### DIFF
--- a/src/browsers.c
+++ b/src/browsers.c
@@ -246,6 +246,7 @@ static const char *browsers[][2] = {
   {"SeznamBot", "Crawlers"},
   {"coccocbot", "Crawlers"},
   {"Neevabot", "Crawlers"},
+  {"Mail.RU_Bot", "Crawlers"},
 
 
   /* Podcast fetchers */
--- a/src/browsers.c
+++ b/src/browsers.c
@@ -242,6 +242,7 @@ static const char *browsers[][2] = {
   {"MauiBot", "Crawlers"},
   {"Cloud", "Crawlers"},
   {"stagefright", "Crawlers"},
+  {"DataForSeoBot", "Crawlers"},
 
 
   /* Podcast fetchers */
--- a/src/browsers.c
+++ b/src/browsers.c
@@ -247,6 +247,7 @@ static const char *browsers[][2] = {
   {"coccocbot", "Crawlers"},
   {"Neevabot", "Crawlers"},
   {"Mail.RU_Bot", "Crawlers"},
+  {"CCBot", "Crawlers"},
 
 
   /* Podcast fetchers */
--- a/src/browsers.c
+++ b/src/browsers.c
@@ -243,6 +243,7 @@ static const char *browsers[][2] = {
   {"Cloud", "Crawlers"},
   {"stagefright", "Crawlers"},
   {"DataForSeoBot", "Crawlers"},
+  {"SeznamBot", "Crawlers"},
 
 
   /* Podcast fetchers */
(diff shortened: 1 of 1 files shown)


allinurl:
Thanks for submitting this. Actually this caught a regression on the code that checks for bots. I've pushed some changes to development that should catch a lot more bots without the need to explicitly adding them to the list (especially since the size of the list is proportional to the run time). Could I ask you to please build from [**development**](https://github.com/allinurl/goaccess#build-from-github-development) and let me know if those changes do in fact catch those bots you posted above? Thanks.

HannesHofer:
Hi,
When building from master branch most of those bots, the most active in my personal experience beeing DataForSeoBot, are classified as Others. (User Agent beeing Mozilla/5.0).
After applying the changeset and using --ignore-crawlers those seem to be correctly ignored.

But I definitly get your concern in maintaining a ever-growing list of bots.
Do you think generally classifing all user agents containing "Bot" as Crawler would be a feasible/better solution?

allinurl:
Good point, I thought about that, I guess the tricky part would be trying to get the actual name of the bot as opposed to displaying the entire UA. I noticed that most of the legit bots contain an +http address and thus I added the [**functionality**](https://github.com/allinurl/goaccess/blob/master/src/browsers.c#L473) to extract that, but again there're a bunch of other bots that don't. You could try outputting `--unknowns-log=<filename>` to see a list of unknowns and check what type of traffic you are getting.

Also I'd like to go over the [**default list**](https://github.com/allinurl/goaccess/blob/master/src/browsers.c#L192) of crawlers and see which ones are covered by the http extraction code and reduce the size of that list as it should make goaccess faster when it comes to parsing the log.

HannesHofer:
What would you think to treat all UA containing URI-schemes 'http://' and 'https://' as Crawlers?
changing `check_http_crawler` to:
```
if ((match = strstr (str, "http://")) || (match = strstr (str, "https://")))
     return match;
return NULL;
```

This would cover all the bots I added and possibly all already present.
In my tests this would improve parsing performance by ~28% from ~6seconds to 4.3seconds on my machine with a 320000 entries access.log

I also checked which crawlers would be detected without having a explicit crawler list.
All crawlers were correctly detected, however I'm missing around 35 out ouf the 62 explicitily configured crawlers in my example access.log . I think those would be most likely detected as well, however I cannot test this.
(See list below of missing crawlers)

The changeset for this change would be easy and I can create a new pull request for this if you like.

not covered crawlers in my access.log:
```AdsBot-Google
AppEngine-Google
Mediapartners-Google
AppleBot
Jakarta Commons-HttpClient
netEstate
PiplBot
IstellaBot
heritrix
PagesInventory
rogerbot
fastbot
yacybot
PycURL
AndroidDownloadManager
Embedly
ruby
Ruby
LinkedIn
Microsoft-WebDAV
ZoominfoBot
Netcraft Web Server Survey
Pandalytics
Nimbostratus-Bot
HTTP Banner Detection
Hakai
WinHttp.WinHttpRequest.5
NetSystemsResearch
Nextcloud Server Crawler
GoScraper
ZmEu
DowntimeDetector
MauiBot
Cloud
stagefright
```

allinurl:
I'm a little hesitant just because I'm not sure we can extract the right name if we go that far. However, I'm curious to see how those logs look like. Are you able to post those user agents directly from your log? Thanks

HannesHofer:
Hi,

It seems that I made a mistake on testing.
The "issue" is indeed fixed for all additonal crawlers already.
sorry for the confusion.

allinurl:
Got it. Thanks for letting me know!

PR #2269, opened 2022-01-26
Do not crash on argument counter of zero

cgzones:
Check before calling xstrdup, which does not accept a NULL pointer,
whether argv[0] is actually set.


allinurl:
Merged. Thank you!

PR #2278, opened 2022-02-18
Added the ability to hide panels on report

ramiresviana:
GoAccess provides a lot of information in the reports, which is very useful and serves many users. Therefore, reports contain multiple panels, some of them may not be useful for some users or at a given time. There are already options to control the displayed panels, but they can only be set when generating the report.

This PR allows users to toggle panels on reports at any time by clicking on a icon on the sidebar panel list.

image


allinurl:
Merged. Thanks so much for submitting this PR, this is awesome!

Issue #2283, opened 2022-03-03
"Cleaning up resources..." string prints to stderr.


homeisfar:
https://github.com/allinurl/goaccess/blob/master/src/goaccess.c#L213

Is this intentional? This seems like normal functionality that should print to stdout. As it stands, this extra stderr print causes spurious mail to be sent from a cron job, as this user describes: https://github.com/allinurl/goaccess/issues/467#issuecomment-1029997006

allinurl:
Thanks for pointing that out. You're right, it should be stdout. Please feel free to submit a PR, happy to merge it :)

PR #2287, opened 2022-03-04
Change stderr to stdout on non-error output

homeisfar:
As discussed in https://github.com/allinurl/goaccess/issues/2283


allinurl:
Merged. Thanks very much!

PR #2293, opened 2022-03-15
Removed the remote port portion from the host format.

### LINKED ISSUE #2220: Planned change to Caddy's access logging
This is part question, part "notice".

In https://github.com/caddyserver/caddy/pull/4403, we're planning on splitting up `remote_addr` into two fields, `remote_ip` and `remote_port`. The change is still pending, and the field names may change before merging.

I wanted to ask if it's possible/easy to support both `remote_addr` and `remote_ip` at the same time in GoAccess, so that it wouldn't be dependent on the version of Caddy?

allinurl:
This should adddress CADDY's v2.5.0 change where the remote port has its own
JSON property.

Fixes #2220

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -78,7 +78,7 @@ static const GPreConfLog logs = {
   "%^ %v [%d:%t %^] %h %^\"%r\" %s %^ %b %^ %L %^ \"%R\" \"%u\"", /* Amazon S3 */
 
   /* Caddy JSON */
-  "{ \"ts\": \"%x.%^\", \"request\": { \"remote_addr\": \"%h\", \"proto\":"
+  "{ \"ts\": \"%x.%^\", \"request\": { \"remote_ip\": \"%h\", \"proto\":"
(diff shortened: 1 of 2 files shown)


src/settings.c
```diff
@@ -78,7 +78,7 @@ static const GPreConfLog logs = {
   "%^ %v [%d:%t %^] %h %^\"%r\" %s %^ %b %^ %L %^ \"%R\" \"%u\"", /* Amazon S3 */
 
   /* Caddy JSON */
-  "{ \"ts\": \"%x.%^\", \"request\": { \"remote_addr\": \"%h:%^\", \"proto\":"
+  "{ \"ts\": \"%x.%^\", \"request\": { \"remote_addr\": \"%h\", \"proto\":"
```
francislavoie, line 81: The field has also been renamed to `remote_ip`, and there's also a new `remote_port` as well

francislavoie:
Awesome, forgot to mention it here that v2.5.0 is out. Thanks @allinurl!

allinurl:
@francislavoie no, thank you for letting me know about this change in Caddy. I plan to push out v1.5.7 (including this change) sometime this week.

PR #2318, opened 2022-05-14
Show city with the DBIP City Lite database too.

mikaraunio:
Currently, goaccess will only display host cities if the geoip database type id matches Maxmind's. This relaxes the check to match the DBIP City Lite database as well.


allinurl:
Merged. Thanks a lot for submitting this!

mikaraunio:
Happy to help, thanks for merging.

Issue #2321, opened 2022-05-17
Invalid Ubuntu signatures


jonathanza:
The following command from https://goaccess.io/download#official-repo is failing:

```
sudo apt-get update
Hit:1 http://za.archive.ubuntu.com/ubuntu focal InRelease
Hit:2 http://za.archive.ubuntu.com/ubuntu focal-updates InRelease
Hit:3 http://za.archive.ubuntu.com/ubuntu focal-backports InRelease
Hit:4 http://ppa.launchpad.net/git-core/ppa/ubuntu focal InRelease
Hit:5 http://nginx.org/packages/ubuntu focal InRelease
Hit:6 http://security.ubuntu.com/ubuntu focal-security InRelease
Hit:7 http://ppa.launchpad.net/ondrej/php/ubuntu focal InRelease
Hit:8 https://esm.ubuntu.com/infra/ubuntu focal-infra-security InRelease
Get:9 https://deb.goaccess.io focal InRelease [2,345 B]
Hit:10 https://esm.ubuntu.com/infra/ubuntu focal-infra-updates InRelease
Err:9 https://deb.goaccess.io focal InRelease
  The following signatures were invalid: EXPKEYSIG 97BD1A0133449C3D Gerardo Orellana <goaccess@prosoftcorp.com>
Fetched 2,345 B in 4s (532 B/s)
Reading package lists... Done
W: An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: https://deb.goaccess.io focal InRelease: The following signatures were invalid: EXPKEYSIG 97BD1A0133449C3D Gerardo Orellana <goaccess@prosoftcorp.com>
W: Failed to fetch https://deb.goaccess.io/dists/focal/InRelease  The following signatures were invalid: EXPKEYSIG 97BD1A0133449C3D Gerardo Orellana <goaccess@prosoftcorp.com>
W: Some index files failed to download. They have been ignored, or old ones used instead.
```

I downloaded the key again but the same error occurred.

ahayes:
Thank you for creating this issue. I have been trying to find my way around the same problem for the last 30 minutes or so with no success. It looks like this issue comes up for various projects when the repo signing key changes.

klosterbruder:
I can confirm this problem on Debian as well. The repository key needs to be renewed or replaced.

[...]

PR #2322, opened 2022-05-17
Include arch in ubuntu/debian repo instructions

ahayes:
See https://github.com/allinurl/goaccess/issues/2321#issuecomment-1129074851


allinurl:
Merged. Thank you!

Issue #1663, opened 2020-02-02
Realtime Stats WebSocket Gets Disconnected After 1 Minute of Inactivity


Elepover:
## Symptom
Every after exactly **1 exactly** minute of inactivity, the WebSocket gets disconnected like this:

![image](https://user-images.githubusercontent.com/11826985/73604463-271bce00-45cc-11ea-8c1e-b6af7b1e8c29.png)

On servers with a large amount of requests, the problem may not appear for the stats are always being updated quickly and therefore the WebSocket is always active. But on small servers like this one, there is totally possible to have a 1-minute-no-request period.

There have already been an issue discussing about this problem: #1491, I'll talk about that in the `Suggested Solution` part.

## Analysis

It's not GoAccess's fault.

It may be due to the limitation of Firefox, Chrome and possibly other browsers. They kills WebSocket connections that have been inactive for 1 minute. Reference: [This StackOverflow Post](https://stackoverflow.com/questions/49408031/websockets-in-chrome-and-firefox-disconnecting-after-one-minute-of-inactivity)

Or, it may be nginx's fault for the `proxy_read_timeout`  field.

## Suggested Solution

As is described in #1491, extending `proxy_read_timeout` did work, but it's not the perfect solution since there may be a even longer "no-request period". And extending the period is somehow "bypassing a security feature".

Here are some suggested solutions or "workarounds":

- Implement a _"heartbeat" mechanism_ on the **backend** to keep the WebSocket connection alive. (Which has already been mentioned in #1491)
- Try automatically reconnecting the WebSocket on the **frontend**.

allinurl:
Thanks for reporting this and pointing that out. Like you said (+ [here](https://stackoverflow.com/questions/10585355/sending-websocket-ping-pong-frame-from-browser)), we could use pings to resolve this or just reconnect when you need to use the socket again. I'm more incline for sending a ping/pong frame.

I can look into this.

Chi-Yu:
I ran into the same issue with a low traffic domain and first thought that it was me doing something wrong.

Khez:
Ran into the same issue, ended up adding
```
window.GoAccess.wsKeepAlive = function(socket) {
    // jumpstart the "loop"
    keepAlive();

    // send data after 30s
    function keepAlive() {
        setTimeout(onKeepAlive, 30000);
    }

    // what a great time to keep alive
    function onKeepAlive() {
        // we got disconnected
        // we could reconnect or do nothing
[...]

PR #2339, opened 2022-06-15
Add option --ping-interval=<secs>

mario-donnarumma:
Solves #1663

Some notes:
- there is no way to send "real" ping/pong frames (sections 5.5.2 and 5.5.3 of [RFC 6455](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.2)) with the JavaScript WebSocket API - further discussion [here](https://stackoverflow.com/questions/10585355/sending-websocket-ping-pong-frame-from-browser)

- the implemented ping mechanism consists in sending plain text from the client-side JavaScript WebSocket; this is because [gwsocket.io](https://gwsocket.io/) (server-side) already supports standardized ping/pong frames as shown below:

  https://github.com/allinurl/goaccess/blob/b30282947bacff987407a71aeb4c5f7a47e6cdf3/src/websocket.h#L151-L152

  https://github.com/allinurl/goaccess/blob/b30282947bacff987407a71aeb4c5f7a47e6cdf3/src/websocket.c#L1912-L1919

  so I thought it was best not to tweak this behavior and simply play around it, in case in the future it becomes possible to send standardized ping/pong frames with the JavaScript WebSocket API

### DIFF
--- a/goaccess.1
+++ b/goaccess.1
@@ -443,6 +443,9 @@ By default, it will attempt to connect to the generated report's hostname. If
 GoAccess is running on a remote server, the host of the remote server should be
 specified here. Also, make sure it is a valid host and NOT an http address.
 .TP
+\fB\-\-ping-interval=<secs>
+Enable WebSocket ping with specified interval in seconds.
+.TP
 \fB\-\-fifo-in=<path/file>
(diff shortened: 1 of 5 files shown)


allinurl:
Awesome, thanks for submitting this PR. Merged!

PR #2375, opened 2022-09-06
Korean PO file added

dcyang:
I just discovered goaccess yesterday and saw that no one seems to have done Korean i18n yet.
I hope this helps.


allinurl:
Merged. 

Thanks so much for this contribution! Much appreciated.

PR #2457, opened 2023-02-04
🐳 Build docker image on different architectures

cristianpb:
Today docker hub registry contains only amd64 architecture.

This PR modifies docker publish script in order to have multiple architectures available. In this case arm64 and amd64.

The image is published only when there is a tag on the master branch.

4 labels are published:
* latest
* version (ex. 1.7.1)
* major.version (ex. 1.7)
* major (ex. 1)

I have tested this in my fork [hub.docker.com/r/cristianpb/goaccess/tags](https://hub.docker.com/r/cristianpb/goaccess/tags)

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -45,30 +45,32 @@ jobs:
     if: github.event_name == 'push'
 
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


allinurl:
Merged. Thank you for submitting this PR!

PR #2481, opened 2023-03-02
Fix spelling errors as reported by the spellcheck action

assistcontrol:
Only comments and display strings are changed here.

Notable here is the correction of referer => referrer. "Referrer" is the actual spelling; "referer" was a misspelling that made it into RFC 1945 (see 1).

It makes the most sense to use "referer" when referring to the HTTP header, and "referrer" when referring to an actual referring page.

[1]: https://en.wikipedia.org/wiki/HTTP_referer


allinurl:
Merged. Thank you!

PR #2520, opened 2023-06-07
enhance(browsers): Updated browsers.c user agents

jordantrizz:
Added a bunch of unknown user agents, as well as some category suggestions.

### DIFF
--- a/src/browsers.c
+++ b/src/browsers.c
@@ -133,7 +133,7 @@ static const char *browsers[][2] = {
   {"aria2", "Others"},
 
   /* WordPress Cron*/
-  {"WordPress/","Cron"},
+  {"WordPress/", "Cron"},
 
   /* Feed-reader-as-a-service */
   {"AppleNewsBot", "Feeds"},


src/browsers.c
```diff
@@ -49,7 +49,7 @@ static char ***browsers_hash = NULL;
 
 /* {"search string", "belongs to"} */
 static const char *browsers[][2] = {
-  /* Game systems: most of them are based of major browsers,
+  /* Game systems: most of them are based off major browsers,
    * thus they need to go before. */
```
Sea-n, line 53: I think this might be "based on" instead of "based of" or "based off".

The original sentence is:
> Game systems: some of them are Safari-based, thus they need to go before Safari.
> Ref: 0e9dc00 (2015 June)

src/browsers.c
```diff
@@ -132,6 +132,9 @@ static const char *browsers[][2] = {
   {"libmpv", "Others"},
   {"aria2", "Others"},
 
+  /* WordPress Cron*/
+  {"WordPress/","Cron"},
```
Sea-n, line 136: It would be better to have a space after comma, please check your following lines as well.

```suggestion
  {"WordPress/", "Cron"},
```

PR #2561, opened 2023-09-20
Macintosh -> macOS

sveinbjornt:
Referring to macOS (previously OS X, before that Mac OS X) as "Macintosh" is archaic terminology. Fixed that.


allinurl:
Thanks for the heads up; the change has been merged.

PR #2588, opened 2023-12-05
Optimize include stmts, Fix IDE warnings and typos

Sea-n:
This PR mainly does the following things:
- Removes unused `#include <xxx.h>` statements and adds back the needed ones
- Fixes some typos in comments, source code, and documents
- Fixes some warnings flagged by my IDE (CLion by JetBrains)

The commits are split into small steps. Please let me know if any should be broken down further.
And this could be done through several PRs if it's more convenient for you. (Please advise on how to break it down.)

The code has been tested on my Ubuntu 22.04 and macOS environments, but I cannot guarantee about the edge cases (e.g., `-Dmarcos`).

---

BTW, thanks for the quick mail reply 5 weeks ago! 😍
We are still working on the parallelize problem, hope we can bring the result soon.


allinurl:
This is great — merged. Thanks very much!

PR #2591, opened 2023-12-09
Depreacte --output-format, only allow --output option

Sea-n:
### Current situation
- The help message says "`-o --output=file.html|json|csv` - Output either an HTML, JSON or a CSV file."
- While using `-o` (and the deprecated `--output-format`), it will trigger [read_option_args: case 'o'](https://github.com/allinurl/goaccess/blob/v1.8/src/options.c#L810-L815)
    - This one checks `valid_output_type` first, then stores it to `output_formats`.
- When using `--output`, it will trigger [parse_long_opt: "output"](https://github.com/allinurl/goaccess/blob/v1.8/src/options.c#L555-L557)
    - This one stores it to `output_formats` directly.

### The bug I encountered
While using `--output out.tsv`, it runs through the parsing process and has no output file.
When changed to `-o out.tsv` for short, it finally gives me the "invalid filename extension" error.

### Refer to commit history
- In 2016 Jun 2e7c4b8, Added an `--output=file.[html|csv|json]` as a shortcut to `--output-format`
- In 2017 Apr 91da175, Updated README.md with Docker details
    - Changed `output-format` to `output` in README.
- In 2017 Jun 6309ed6, Fix error in example config file
    - The latest `--output-format` in this repo (except `src/options.c`) is removed.
- In 2017 Oct 27d8782, Add `valid_output_type()` to validate the filename given via `-o`
    - The file extension check is only applied for `-o` (`--output-format`), and the PR author forgot about `--output`.

According to the commit history, we can see that `--output-format` has been replaced for 6+ years. I think it's safe to remove it without any warning.

So after this PR is merged, `--output-format` will only output a fatal error, saying it's deprecated, without backward compatibility.


src/options.c
```diff
@@ -263,7 +263,8 @@ cmd_help (void)
   "                                    ranges. e.g., 192.168.0.1-192.168.0.10\n"
   "  -H --http-protocol=<yes|no>     - Set/unset HTTP request protocol if found.\n"
   "  -M --http-method=<yes|no>       - Set/unset HTTP request method if found.\n"
-  "  -o --output=file.html|json|csv  - Output either an HTML, JSON or a CSV file.\n"
+  "  -o --output=<filename|format>   - Output to stdout or the specificed file.\n"
+  "                                    Currently support HTML, JSON and CSV format.\n"
```
Sea-n, line 267: Update:
Now I know `-o file.csv` will save the output to a file, while `-o csv` will directly display it in the terminal.
It's currently not included in the help message. 😅

I need your assistance in improving the proposed wording, thanks. 🙇‍♂️

allinurl: I believe we can include a third line to show both options, and a brief example, like the one you posted above, would suffice. Thanks!

Sea-n:
I still not confident about how to write the text with short word.

Here is some version (with inspiration of [ChatGPT](https://chat.openai.com/share/441cd070-eed7-4ebc-9c16-5dd3c23b2208)), which one would you preferred?

```
(A)
-o --output=<format|filename>   - Output to stdout or the specificed file.
                                  Currently support HTML, JSON and CSV format.
                                                                               *
(B)                                                                            *
-o --output=<format|filename>   - Specify the output format or output filename.
                                  The filename must have .html, .json, or .csv extension.
                                                                               *
(C)                                                                            *
-o --output=<html|json|csv>     - Specify the output format.
-o --output=<FILENAME>          - Save report to a file. The FILENAME must
                                  have .html, .json, or .csv extension.
                                                                               *
(D)                                                                            *
-o --output=[FILE.]<FORMAT>     - Output to stdout or the specificed file.
                                  Supported FORMAT: html, json, csv.
                                                                               *
(E)                                                                            *
-o --output=<html|json|csv|FILE> - Set output format or write to a file.
                                   The FILE must have .html, .json, or .csv extension.
                                                                               *
(F)                                                                            *
-o --output=[file.]<html|json|csv> - Output to stdout or the specificed file.

                                                                               *
(G)                                                                            *
-o --output=<format|filename> - Output to stdout or the specificed file.
                                e.g., --output=report.html, -o out.json, --output=csv
                                                                               *
  (Current longest text)                                                       *
--unknowns-as-crawlers          - Classify unknown OS and browsers as crawlers.
--external-assets               - Output HTML assets to external JS/CSS files.
```

allinurl:
Thanks for posting these options. Let's roll with G; I think that's pretty clear, in my opinion.

Sea-n:
Okay! Just pushed a new version. ☺️
image

And I do a little adjustment here, think it would be more readable.
```
(From this)
-o --output=<format|filename> - Output to stdout or the specified file.
                                e.g., --output=report.html, -o out.json, --output=csv

(Changed to this)
-o --output=<format|filename>   - Output to stdout or the specified file.
                                  e.g.: -o csv, -o out.json, --output=report.html
```

allinurl:
Thanks, Sean! The changes have been merged.

PR #2598, opened 2023-12-21
Add newline to HTML output

Sea-n:
The current version prints all HTML elements in one line

Current Version

Add line break for every elements.

New Version

This only add 60 bytes to ~500 KB file.
But may improve the user experience when open the source code.

Also added "GoAccess 1.x" and GitHub Repo URI in last line.


allinurl:
Thanks for submitting this. I wonder if we can do this conditionally, meaning to include the `\n` based on a compile flag like DEBUG, aiming to avoid repetitive `ifdef` statements for each line and minimizing code duplication. e.g.,

```
#ifdef _DEBUG
  #define NEWLINE "\n"
#else
  #define NEWLINE ""
#endif

static void
print_html_header_styles(FILE *fp, FILE *fcs) {
  if (fcs) {
    fprintf(fp, "<link rel='stylesheet' href='%s'>" NEWLINE, FILENAME_CSS);
    fprintf(fcs, "%.*s" NEWLINE, fa_css_length, fa_css);
    fprintf(fcs, "%.*s" NEWLINE, bootstrap_css_length, bootstrap_css);
    fprintf(fcs, "%.*s" NEWLINE, app_css_length, app_css);
  } else {
    fprintf(fp, "<style>%.*s</style>" NEWLINE, fa_css_length, fa_css);
    fprintf(fp, "<style>%.*s</style>" NEWLINE, bootstrap_css_length, bootstrap_css);
    fprintf(fp, "<style>%.*s</style>" NEWLINE, app_css_length, app_css);
  }
}
```

Sea-n:
I think it would be better to also including newlines for released versions by default.

The less than 1% overhead should not be the problem.
Maybe we can add a flag to remove newlines?

allinurl:
I'm attempting to grasp the underlying concept here. Wouldn't it be simpler to access the developer tools (Ctrl + Shift + C) and view a more neatly formatted output there? I understand it's just a 1% overhead, but I'm uncertain about the distinction between using Control + Shift + C versus Ctrl + U. Let me know. Thanks

Sea-n:
It's just my personal preference maybe.
Close as not needed, thanks for review though 😊

PR #2605, opened 2024-01-05
handle SIGQUIT as a signal for a graceful shutdown

jkittner:
nginx uses `SIGQUIT` as the signal for a graceful shutdown (https://docs.nginx.com/nginx/admin-guide/basic-functionality/runtime-control/)

some context:
- https://github.com/nginxinc/docker-nginx/issues/377
- https://github.com/nginxinc/docker-nginx/issues/167
- https://trac.nginx.org/nginx/ticket/753

In their Dockerfile they also use 

[`STOPSIGNAL SIGQUIT`](https://github.com/nginxinc/docker-nginx/blob/4bf0763f4977fff7e9648add59e0540088f3ca9f/mainline/debian/Dockerfile#L117)

when also running `goaccess` in the same container via [`dumb-init`](https://github.com/Yelp/dumb-init) or something similar, goaccess will exit non zero upon receiving `SIGQUIT` hence the container shutdown will also have a non zero exit code and tools like docker-compose or an `ansible`-based deployment will treat those as errors and abort the process etc..

This patch simply adds `SIGQUIT` as an allowed signal for a graceful shutdown. `SIGKILL` can still be used for non-graceful shutdowns.

Happy to hear your thoughts on this.

### DIFF
--- a/src/goaccess.c
+++ b/src/goaccess.c
@@ -1444,6 +1444,8 @@ handle_signal_action (GO_UNUSED int sig_number) {
     fprintf (stderr, "\nSIGINT caught!\n");
   else if (sig_number == SIGTERM)
     fprintf (stderr, "\nSIGTERM caught!\n");
+  else if (sig_number == SIGQUIT)
+    fprintf (stderr, "\nSIGQUIT caught!\n");
   else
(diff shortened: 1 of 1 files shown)


allinurl:
Merged. Thank you for submitting the pull request!

PR #2621, opened 2024-01-30
Minor fixes in German locale

realpixelcode:
Corrected a few minor spelling, grammar and other mistakes. A couple of notes for other translators:

- In German, the short dash `-` (Bindestrich) is used for composite nouns (multiple dashes for multiple nouns within a single composite word), whereas the longer one `–` (Gedankenstrich) is used for interjections in sentences.
- After the Gedankenstrich `–`, the next character is always in lower-case (except if the word is a noun, obviously).
- Verbs used as nouns are capitalised.

Replaced the English term “panels” with “Kacheln” which means “tiles”, for stylistic reasons; hope that's okay.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -59,11 +59,11 @@ msgstr "Übersicht"
 
 #: src/labels.h:62
 msgid "Dashboard - Overall Analyzed Requests"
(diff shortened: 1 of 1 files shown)


allinurl:
Merged. Thank you so much for submitting this pull request!

Issue #2601, opened 2023-12-23
Caddy log format - client_ip should be used instead of remote_ip


fliespl:
Currently goaccess uses remote_ip in Caddy json format, but according to docs client_ip should be used since it's the one being set with proper ips when using proxies.

https://caddyserver.com/docs/caddyfile/matchers#client-ip

allinurl:
Hey, thanks for letting me know! Do you mind sharing a few lines from your access log so I can check out the latest changes?

fliespl:
Direct access without loadbalancer / proxy:
```
{"level":"info","ts":1703415994.3220897,"logger":"http.log.access.log1","msg":"handled request","request":{"remote_ip":"X.X.X.X","remote_port":"49570","client_ip":"X.X.X.X","proto":"HTTP/1.1","method":"GET","host":"Y.Y,Y,Y:8080","uri":"/index.php","headers":{"Upgrade-Insecure-Requests":["1"],"User-Agent":["Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36"],"Accept":["text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7"],"Accept-Encoding":["gzip, deflate"],"Accept-Language":["pl-PL,pl;q=0.9,en-US;q=0.8,en;q=0.7,fr;q=0.6,es;q=0.5,it;q=0.4,ar;q=0.3,zh-CN;q=0.2,zh;q=0.1,da;q=0.1,sv;q=0.1,nl;q=0.1,fi;q=0.1,de;q=0.1,el;q=0.1,id;q=0.1,he;q=0.1,ja;q=0.1,ms;q=0.1,ru;q=0.1,pt;q=0.1"],"Connection":["keep-alive"],"Dnt":["1"]}},"bytes_read":0,"user_id":"","duration":0.000010499,"size":0,"status":0,"resp_headers":{"Server":["Caddy"]}}
```

client_ip = remote_ip = my public IP

Using loadbalancer / proxy:

```
[...]

PR #2629, opened 2024-02-08
Use `client_ip` for Caddy JSON, allowing trusted proxies

Teqed:
Caddy has added [first-class support for determining client IP](https://github.com/caddyserver/caddy/pull/5104), which means that from Caddy 2.7 beta 1 onwards, the logs include `client_ip` alongside `remote_ip` in the access logs. The value falls back to `remote_ip`'s value, meaning this has no impact on Caddy configurations that don't use trusted proxies.

> the client_ip should always be set (as long as the RemoteAddr is a valid IP address). So in the logs, it will be identical to the remote_ip unless trusted proxies is configured and the proxy passed the client IP in a trusted header.

If trusted proxies are [configured globally](https://caddyserver.com/docs/caddyfile/options#trusted-proxies), Caddy will follow your configuration for determining which proxies are trusted and what headers to use. For example, this allows you to forward traffic from one of your servers to another while trusting IP addresses in the private range.

### DIFF
--- a/config/goaccess.conf
+++ b/config/goaccess.conf
@@ -113,7 +113,7 @@
 #log-format %^ %^ [%h] %^ %^ [%d:%t %^] "%r" %s %b "%R" "%u" %^ %^ [%v] %^:%^ %^ %T %^ %^
 
 # CADDY JSON Structured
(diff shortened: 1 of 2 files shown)


Teqed:
Closes #2601

allinurl:
Merged. Big thanks for sending in this pull request!

PR #2632, opened 2024-02-12
Update sv.po

tellustheguru:
Adjusted some translations and added some


allinurl:
Merged. Thanks very much!

PR #2647, opened 2024-03-18
update ncurses dependency for Ubuntu/Debian

meonkeys:
As of (at least) Ubuntu 22.04 LTS and Debian Buster, there is no longer a `w6-dev` package. The build works as expected with `libncurses-dev` instead.

This is a quick fix only based on the latest releases to keep the readme simple. Please reject this patch if you intend to maintain docs for older Ubuntu/Debian versions.

I tested in a `ubuntu:22.04` container. I installed these packages prior to building:

* autoconf
* automake
* autopoint
* gcc
* gettext
* libncurses-dev
* libgeoip-dev
* libmaxminddb-dev
* libssl-dev
* build-essential


allinurl:
Merged. Thank you!

PR #2679, opened 2024-05-08
Miscellaneous bug fixes and improvements

cgzones:
* Constify read-only data
 
  Declare read-only data arrays const, to warn on accidental modification and let compilers place them in read-only memory sections.

* De-duplicate strtoupper()

  Avoid the duplicate definition of strtoupper() in websocket.c, just use the one from util.h.

* Avoid UB in character checking functions

  The functions classifying characters like isspace(3) require the input to be of type unsigned char or EOF.

* Avoid out-of-bounds access in serial processing mode

  When using just one job for processing the input lines do not use the out-of-bounds index 1 instead of the valid one 0.
  This currently seems to only work since the stack space of the in this case unused threads array is used.

* Insert the status code in the correct position

* Avoid out-of-bounds access for invalid HTTP status codes

  Avoid array access in case the status code is less than 0 or bigger than 599.

  Also reduce the size of the codes array.

* Avoid unnecessary string copies

  Do not duplicate the strings for enumeration keys, as they are read-only.

### DIFF
--- a/src/commons.c
+++ b/src/commons.c
@@ -196,13 +196,13 @@ str2enum (const GEnum map[], int len, const char *str) {
  *
  * On error, -1 is returned.
  * On success, the enumerated module value is returned. */
-char *
+const char *
 enum2str (const GEnum map[], int len, int idx) {
   int i;
 
   for (i = 0; i < len; ++i) {
     if (idx == map[i].idx)
-      return xstrdup (map[i].str);
+      return map[i].str;
   }
 
   return NULL;
@@ -221,7 +221,7 @@ get_module_enum (const char *str) {
  *
  * On error, NULL is returned.
  * On success, the string module value is returned. */
-char *
+const char *
 get_module_str (GModule module) {
   return enum2str (enum_modules, ARRAY_SIZE (enum_modules), module);
 }
--- a/src/commons.c
+++ b/src/commons.c
@@ -54,7 +54,7 @@ int module_list[TOTAL_MODULES] = {[0 ... TOTAL_MODULES - 1] = -1 };
 
 /* *INDENT-OFF* */
 /* String modules to enumerated modules */
-static GEnum enum_modules[] = {
+static const GEnum enum_modules[] = {
   {"VISITORS"        , VISITORS}        ,
   {"REQUESTS"        , REQUESTS}        ,
   {"REQUESTS_STATIC" , REQUESTS_STATIC} ,
--- a/src/browsers.c
+++ b/src/browsers.c
@@ -47,7 +47,7 @@
 static char ***browsers_hash = NULL;
 
 /* {"search string", "belongs to"} */
-static const char *browsers[][2] = {
+static const char *const browsers[][2] = {
   /* Game systems: most of them are based on major browsers,
    * thus they need to go before. */
   {"Xbox One", "Game Systems"},
--- a/src/color.c
+++ b/src/color.c
@@ -49,7 +49,7 @@ static GSLList *color_list = NULL;
 static GSLList *pair_list = NULL;
 
 /* *INDENT-OFF* */
-static GEnum CSTM_COLORS[] = {
+static const GEnum CSTM_COLORS[] = {
   {"COLOR_MTRC_HITS"              , COLOR_MTRC_HITS},
   {"COLOR_MTRC_VISITORS"          , COLOR_MTRC_VISITORS},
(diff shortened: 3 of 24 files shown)


src/commons.c
```diff
@@ -196,13 +196,13 @@ str2enum (const GEnum map[], int len, const char *str) {
...
  * On success, the enumerated module value is returned. */
-char *
+const char *
 enum2str (const GEnum map[], int len, int idx) {
   int i;
 
   for (i = 0; i < len; ++i) {
     if (idx == map[i].idx)
-      return xstrdup (map[i].str);
+      return map[i].str;
```
allinurl, line 205: I'm getting an invalid pointer without the strdup. Are you seeing this on your end too?  Thanks!

```
./goaccess ../10.log -o a.html --log-format=VCOMBINED
free(): invalid pointer
Aborted (core dumped)
Thread 1 "goaccess" received signal SIGABRT, Aborted.
```

cgzones: No, I am not seeing such memory corruption.
Do you use some non default settings in the global goaccess.conf?
Can you share the stacktrace of the crash or a public version of the triggering input log?
Otherwise the following patch might resolve it?
```diff
---
 src/gkhash.c | 23 ++++++++++++++---------
 1 file changed, 14 insertions(+), 9 deletions(-)

diff --git a/src/gkhash.c b/src/gkhash.c
index d40e942d..3109c78c 100644
--- a/src/gkhash.c
+++ b/src/gkhash.c
@@ -1040,8 +1040,10 @@ inc_su64 (khash_t (su64) *hash, const char *key, uint64_t inc) {
     dupkey = xstrdup (key);
     k = kh_put (su64, hash, dupkey, &ret);
     /* operation failed */
-    if (ret == -1)
+    if (ret == -1) {
+      free(dupkey);
       return -1;
+    }
   } else {
     value = kh_val (hash, k) + inc;
   }
@@ -1087,6 +1089,7 @@ static uint32_t
 inc_si32 (khash_t (si32) *hash, const char *key, uint32_t inc) {
   khint_t k;
   int ret;
+  char *dupkey = NULL;
 
   if (!hash)
     return 0;
@@ -1094,10 +1097,16 @@ inc_si32 (khash_t (si32) *hash, const char *key, uint32_t inc) {
   k = kh_get (si32, hash, key);
   /* key not found, put a new hash with val=0 */
   if (k == kh_end (hash)) {
-    k = kh_put (si32, hash, key, &ret);
+    dupkey = xstrdup(key);
+    k = kh_put (si32, hash, dupkey, &ret);
     /* operation failed */
-    if (ret == -1)
+    if (ret == -1) {
+      free(dupkey);
       return 0;
+    }
+    /* concurrently added */
+    if (ret == 0)
+      free(dupkey);
     kh_val (hash, k) = 0;
   }
 
@@ -1484,9 +1493,7 @@ ht_inc_cnt_overall (const char *key, uint32_t val) {
   if (!hash)
     return 0;
 
-  if (get_si32 (hash, key) != 0)
-    return inc_si32 (hash, key, val);
-  return inc_si32 (hash, xstrdup (key), val);
+  return inc_si32 (hash, key, val);
 }
 
 int
@@ -1509,9 +1516,7 @@ ht_ins_seq (khash_t (si32) *hash, const char *key) {
   if (!hash)
     return 0;
 
-  if (get_si32 (hash, key) != 0)
-    return inc_si32 (hash, key, 1);
-  return inc_si32 (hash, xstrdup (key), 1);
+  return inc_si32 (hash, key, 1);
 }
 
 /* Insert an IP hostname mapped to the corresponding hostname.
-- 
2.43.0
```

allinurl: The patch didn't resolve the issue. However, to reproduce it, you'll need to build with `--enable-debug`. Without this flag, the issue doesn't occur. I'll incorporate some additional tests into the build workflow to help catch future issues. e.g., 

    ./configure --enable-utf8 --enable-geoip=mmdb --enable-debug --with-getline--with-openssl
    echo 'google.com.com:80 192.168.88.1 - - [27/Apr/2024:16:29:29 -0500] "GET / HTTP/1.1" 403 1431 "-" "Mozilla/5.0 (X11 ; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36" 1803' | ./goaccess - --log-format=VCOMBINED

allinurl:
Great, thank you for addressing these issues. Merged!

PR #2681, opened 2024-05-16
Update hash table increment functions

cgzones:
Free the duplicated key in inc_su64() on failure.

Use the same pattern as in inc_su64() for inc_si32(), by duplicating the key after an empty lookup result. This simplifies the call sites ht_inc_cnt_overall() and ht_ins_seq() by avoiding a duplicate lookup.


allinurl:
Merged. Appreciate you sending the pull request!

PR #2682, opened 2024-05-17
Add some dependencies to install and make every shell command copy-pastable

Bbertatum:
When I needed to install goaccess the make step failed because of missing dependencies (maxmind and ncurses), so I added it in README in the style Debian style. 
Then I also wish to have every command copy-pastable without removing the "$" or "#" char so I remove it and add it for the real comments.


Bbertatum:
Last commit is there to include last changes from master branch

Bbertatum:
Last commit is there to include last changes from master branch

README.md
```diff
@@ -101,29 +101,30 @@ GoAccess can be compiled and used on *nix systems.
 
 Download, extract and compile GoAccess with:
 
-    $ wget https://tar.goaccess.io/goaccess-1.9.2.tar.gz
-    $ tar -xzvf goaccess-1.9.2.tar.gz
-    $ cd goaccess-1.9.2/
-    $ ./configure --enable-utf8 --enable-geoip=mmdb
-    $ make
-    # make install
+    sudo apt install libncursesw5-dev libmaxminddb-dev
```
allinurl, line 104:
```suggestion
```

allinurl: I think it's best to keep this part distro-agnostic. We can emphasize the dependencies at the bottom, where they're already [listed](https://github.com/allinurl/goaccess?tab=readme-ov-file#distribution-packages).

Bbertatum: I think that it's not the "straightforward way" to have all this deps in a table like this because per-distro commands to make things work are written above (and also I missed it... thx to point it out) maybe I missed something else but these deps are not "optional" because without it the software do not compile.
But I also agree with your way to see things. Thx for your beautiful software BTW :)

PR #2739, opened 2024-10-20
remove dup DotBot in browsers.list

meonkeys:
fixes `Duplicate browser entry: DotBot` error when this file is used with `browsers-file`


allinurl:
Merged. Thank you for submitting this PR!

PR #2799, opened 2025-03-07
Include stdint.h

NattyNarwhal:
We use this, but it may not be included via other headers; fixes building on AIX.


allinurl:
Merged. Thank you, I appreciate you submitting this PR and glad to hear it runs on AIX :)

PR #2800, opened 2025-03-07
Don't swap bytes on big endian platforms as a fallback

NattyNarwhal:
We're already on the network byte order, there's no need to swap bytes. For instance, AIX doesn't have a header for this, but it's already on a big endian platform, so it's unnecessary.


allinurl:
Merged. Thank you!

PR #2830, opened 2025-04-18
Better Chinese translation

azio7:


allinurl:
Merged, thanks very much for submitting this!

PR #2838, opened 2025-05-06
Make error messages more actionable.

branc116:
Changed src/parser.c so that error messages are more understandable.
I have spend way too long trying to figure out what "IPv4/6 is required" is.

I think it'd be more user friendly to have additional message for a user to tell him what exactly he needs to do.
If I only have that message I think I'd be done with my task way sooner.

P.S.
I like this project once I got it working . I also have a things I wanna change. I look forward to it.
I have also seen a lot of issues in this repository that mention this error and people are stuck, so I think this will also help them, help you because you will no loger get the same issues opend ( hopefully. ( most likely no lol ) )

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -1653,11 +1653,11 @@ static int
 verify_missing_fields (GLogItem *logitem) {
   /* must have the following fields */
   if (logitem->host == NULL)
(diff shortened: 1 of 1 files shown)


allinurl:
Merged. Thank you, that will definitely improve the ux.

PR #2843, opened 2025-05-20
ASN panel mention in goaccess.conf

sh4rkman:
I got confused when i enabled the GEO_LOCATION panel to discover an "ASN Panel" that wasn't mentioned anywhere in the goaccess.conf parameters list.

This pull request updates the `config/goaccess.conf` file to include the ASN (Autonomous System Number) panel in both enabled and ignored panel configurations.

Changes to `config/goaccess.conf`:

* Added `#enable-panel ASN` to the list of enabled panels.
* Added `#ignore-panel ASN` to the list of ignored panels.


allinurl:
Merged. Thank you for catching that!

Issue #2848, opened 2025-06-03
Serving HTML report to browsers with `--real-time-html` server


NattyNarwhal:
One pain point we've been having with using GoAccess on remote servers is there's a few moving parts to serve the real-time HTML report. We need a web server running to serve the HTML report (sometimes you can reuse an existing one, but this might not necessarily be the case), and obviously it needs GoAccess running for the WebSocket.

For cases like this where you want to view the report from a remote system, in a browser with minimal fuss, I'm thinking it might be possible to use the existing WebSocket server for this - just serve the HTML report instead of only serving a WebSocket. This would make it easier to be able to serve it with just a single command to start GoAccess (and make it easier to manage if you want to leave it running too).

I'm currently implementing this, though it has some problems - i.e. sometimes the existing socket send for WebSockets will drop, likely because it wasn't designed for sending whole files. If this is a feature that you want upstream, let me know.

allinurl:
That sounds like a cool idea, sending the HTML report over a WebSocket. but it seems like you'd need a client (maybe the report itself?) hosted on a web server to serve the page and open a connection to the GoAccess WebSocket server?

NattyNarwhal:
WebSockets work over HTTP, so what could be done is instead of failing on any requests that don't upgrade to WebSockets, it could serve the HTML report back, since most of the work of the web server is there.

Since you're interested, I'll file a PR soon.

PR #2849, opened 2025-06-04
Use WebSocket HTTP server to serve HTML reports

NattyNarwhal:
See #2848 for the justification, but it should simplify the case of viewing the HTML reports when GoAccess is being run on another system.

On mind for this:

- The send code is somewhat unreliable; it will stop accepting new data for the buffers to send, truncating the output.
- ~~Doesn't handle static resources split out to other files yet (`--external-assets`)~~ Does so now.
- ~~It'd probably be more efficient to read the file once and then send it, since the report shouldn't change on disk until it's re-ran. It might help alleviate the issues with ~~ Doesn't seem to help this part - though it seems to truncate around ~32k, and small reports/the CSS are fine, larger reports, the JS are not.
- It may be better to make `find_output_type` not allocate at all and make callers do that.

### DIFF
--- a/src/gwsocket.c
+++ b/src/gwsocket.c
@@ -45,6 +45,7 @@
 #include "settings.h"
 #include "websocket.h"
 #include "wsauth.h"
+#include "util.h"
 #include "xmalloc.h"
 
 /* Allocate memory for a new GWSReader instance.
@@ -488,6 +489,7 @@ start_server (void *ptr_data) {
 /* Read and set the WebSocket config options. */
 static void
 set_ws_opts (void) {
+  const char *html_filename;
   ws_set_config_strict (1);
   if (conf.addr)
     ws_set_config_host (conf.addr);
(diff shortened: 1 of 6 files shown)


allinurl:
Thanks for this PR! I'm seeing a `net::ERR_CONTENT_LENGTH_MISMATCH` response, and as you mentioned, it seems like `ws_respond` is only sending 108,600 bytes instead of the expected 734,919 bytes (file size).

Since the WS server is fundamentally built around the event loop, sending the remainder of the HTTP response will need to integrate with it properly. Given that the response is being queued in the send buffer, we'll need to adjust how `ws_send_http_response` is invoked and to use the event loop. I'll take a close look and follow up.

NattyNarwhal:
I've rebased this and fixed the issue with incomplete sends (was cache buffers not getting sent with incomplete send(2)), but I'm not entirely confident that it's the right place to drain the queue for that.

PR #2872, opened 2025-08-04
Add Alibaba Cloud Linux

metalefty:
See: https://www.alibabacloud.com/en/product/alibaba-cloud-linux-2

Here is an example of the user-agent:
    libdnf (Alibaba Cloud Linux 3; generic; Linux.x86_64)


allinurl:
Merged. Thank you!

PR #2875, opened 2025-08-18
Update sv.po

tellustheguru:
Swedish translation updated

### DIFF
--- a/po/sv.po
+++ b/po/sv.po
@@ -974,7 +974,8 @@ msgid "505 - HTTP Version Not Supported"
 msgstr "505 - HTTP-version stöds inte"
 
 #: src/labels.h:561
-msgid "509 - Bandbreddsgräns överskriden: Servern har överskridit bandbredden"
+msgid "509 - Bandwidth Limit Exceeded: The server has exceeded the bandwidth"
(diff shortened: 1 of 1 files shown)


po/sv.po
```diff
@@ -1002,17 +967,14 @@ msgstr "503 - Service Otillgänglig: Servern är för närvarande inte tillgäng
...
+msgstr "504 - Gateway Timeout: Uppströms-servern misslyckades med att skicka förfrågan"
 
 #: src/labels.h:559
 msgid "505 - HTTP Version Not Supported"
 msgstr "505 - HTTP-version stöds inte"
 
 #: src/labels.h:561
-msgid "509 - Bandwidth Limit Exceeded: The server has exceeded the bandwidth"
-msgstr ""
+msgid "509 - Bandbreddsgräns överskriden: Servern har överskridit bandbredden"
```
allinurl, line 977: Thanks for the update. It seems like at line 977 it has a msgid but is missing the corresponding msgstr. This breaks msgmerge, which is used to update translations during the build.

tellustheguru:
Fix the error...

allinurl:
Merged. Thank you!

PR #2892, opened 2025-10-25
Add vhost to request string

JoshMBriggs:
I couldn't find an easy way to process multiple files from Nginx Proxy Manager and have the request string include the vhost.
I've added an argument to concatenate the vhost to the request string.


allinurl:
Merged. Thank you!

PR #2928, opened 2026-04-08
Include `--with-zlib` and `--enable-asan` in version info when configured

ChandlerSwift:
I did also check for other missing options, and these are the only two options documented at https://goaccess.io/man#configuration but not included in the output.


allinurl:
Merged. Thank you!

PR #2947, opened 2026-05-21
fix: add large file support defines to fileio.c and persistence.c

sithglan:
Without _FILE_OFFSET_BITS=64 and related LFS defines, fopen() and stat() fail with EOVERFLOW on 32-bit systems when opening log files larger than 2GB. Every other source file in the project already defines these macros.

The issue we had on 32 bit Debian Trixie:
Unable to open the specified log file '/var/log/nginx/access.log.1'. Value too large for defined data type

### DIFF
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -29,6 +29,10 @@
  * SOFTWARE.
  */
 
+#define _LARGEFILE_SOURCE
+#define _LARGEFILE64_SOURCE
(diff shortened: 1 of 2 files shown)


allinurl:
Thanks for the PR. Could you please rebase this on the latest master and push again?

sithglan:
I rebased on master and pushed again.

PR #2951, opened 2026-05-25
Add persisted database retention window

TheVillageGuy:
Adds `--persist-retention=<days>[d]` to limit how much data is restored from a persisted database.

When GoAccess is run with `--restore --persist --persist-retention=365d`, it restores the existing persisted database but ignores date buckets older than the retention window. When the database is persisted again, only the retained date buckets are written back.

This is useful for large persisted databases where older aggregate data is no longer needed, but recent analytics should be preserved.

Example:

```bash
goaccess access.log --restore --persist --db-path /path/to/db --persist-retention=365d
```

### DIFF
--- a/config/goaccess.conf
+++ b/config/goaccess.conf
@@ -593,6 +593,11 @@ ignore-panel KEYPHRASES
 #
 # keep-last 7
 
+# Retain only persisted storage entries newer than now minus the given number
+# of days when restoring. The suffix 'd' is optional.
+#
+# persist-retention 365d
+
 # Disable client IP validation. Useful if IP addresses have been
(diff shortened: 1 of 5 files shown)


allinurl:
Thanks for submitting this PR. I'm trying to understand it a bit better, how does this help compared to just using `--keep-last=<ndays>`?

TheVillageGuy:
This PR is useless, the functionality is already in there, I just completely missed it. Sorry for wasting time, including my own.

allinurl:
Oh, no problem at all, I really appreciate the contribution, though. I just wanted to make sure we weren't introducing duplicate functionality.

PR #2956, opened 2026-06-08
minor fix: Updated man page to prevent .gz from being interpreted as a macro

gemmaro:
In the `--with-zlib` option description, the line wrapped such that `.gz` appeared at the start of a line, which causes its misinterpretation as a macro directive.  This fixes by reflowing the sentence.


allinurl:
Merged. Thanks very much!

PR #2970, opened 2026-07-15
Add QUERY to the list of http methods in gstorage.c

torstehu:
RFC 10008[1] was published 15. June 2026, standardizing the use of HTTP QUERY as a way for safe, idempotent and cachable http calls.

Refs:
 - [1] https://www.rfc-editor.org/info/rfc10008


allinurl:
Merged. Thank you!

PR #2975, opened 2026-08-03
Show hostname resolution progress for static reports

ElliottLandsborough:
# Summary

When generating static reports (HTML/JSON/CSV via --with-output-resolver), GoAccess reverse-resolves the hostname for every visitor IP. On large logs this can take a while, and previously there was no feedback — the parsing spinner finished, then the process just appeared to hang while DNS lookups happened silently.

This PR adds a second, determinate progress phase to the existing spinner: after parsing completes, hostname resolution now shows Resolving hostnames X/Y (Z pending) so users can see it is actively working and roughly how far along it is.

Screen recording of it switching over from processing 70 million logs to resolving the hostnames:

https://github.com/user-attachments/assets/96241b53-c92d-4af8-a008-17a320038cc3

# Testing

 - Manually ran against large amount of logs logs with --with-output-resolver; confirmed hostnames resolve correctly in the output report and the progress counter advances and matches the final resolved count.
 - No change in behaviour for real-time HTML / curses mode (progress display remains parsing-only there, as before).


allinurl:
Merged, thank you! I also applied a sync fix on top of it.

PR #2976, opened 2026-08-04
Use %zu printf format for type size_t

edgar-bonet:
I am still using a 32-bit netbook, and even compiling stuff on it. Yes, I know it's outdated. While building goaccess, I got this GCC warning:

```
src/bin2c.c:119:39: warning: format ‘%lu’ expects argument of type ‘long unsigned int’,
but argument 4 has type ‘size_t’ {aka ‘unsigned int’} [-Wformat=]
  119 |   fprintf (f_output, "const char %s[%lu] = {", ident, file_size);
      |                                     ~~^               ~~~~~~~~~
      |                                       |               |
      |                                       |               size_t {aka unsigned int}
      |                                       long unsigned int
      |                                     %u
```

and the same warning on line 137.

This pull request fixes the printf format, which should be `%zu`.

### DIFF
--- a/src/bin2c.c
+++ b/src/bin2c.c
@@ -116,7 +116,7 @@ main (int argc, char *argv[]) {
   ident = argv[3];
   need_comma = 0;
 
-  fprintf (f_output, "const char %s[%lu] = {", ident, file_size);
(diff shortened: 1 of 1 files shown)


allinurl:
Awesome. Merged. Thank you!

